PR #5, opened 2016-08-26
Pivotal Plugin

kjlundsgaard:
@getsentry/infrastructure @macqueen

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/5)


src/sentry_plugins/pivotal/plugin.py
```diff
@@ -0,0 +1,149 @@
...
+from rest_framework.response import Response
+from django import forms
+from django.conf.urls import url
+from django.utils.translation import ugettext_lazy as _
+from django.utils.encoding import force_text
+from sentry.plugins.bases.issue2 import IssuePlugin2, IssueGroupActionEndpoint
+from sentry.utils.http import absolute_uri
+
+import sentry_plugins
+import pyvotal
```
mattrobenolt: This definitely needs to be added into `setup.py` as a dependency.

src/sentry_plugins/pivotal/plugin.py
```diff
@@ -0,0 +1,149 @@
...
+            'name': 'issue_id',
+            'label': 'Story',
+            'default': '',
+            'type': 'select',
+            'has_autocomplete': True,
+            'help': ('Search Pivotal Stories by name or description.')
+        }, {
+            'name': 'comment',
+            'label': 'Comment',
+            'default': absolute_uri(group.get_absolute_url()),
```
mattrobenolt: This `absolute_uri` is redundant here since `group.get_absolute_url()` is doing that.

src/sentry_plugins/pivotal/plugin.py
```diff
@@ -0,0 +1,149 @@
...
+        if not comment:
+            return
+        client = self.get_api_client(group.project)
+        try:
+            project = client.projects.get(self.get_option('project', group.project))
+            story = project.stories.get(form_data['issue_id'])
+            story.add_note(comment)
+        except pyvotal.exceptions.AccessDenied:
+            raise forms.ValidationError(_('Access denied through Pivotal API.'))
+        except Exception, e:
```
mattrobenolt: `except Exception as e` plz

macqueen:
*[src/sentry_plugins/pivotal/plugin.py, line 97 \[r2\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQNWrgK1evkn7nIE6aU:-KQNWrgK1evkn7nIE6aV:b-wy561e) ([raw file](https://github.com/getsentry/sentry-plugins/blob/2e29158ee7ad04f3d5eed5854d39774f86832a5b/src/sentry_plugins/pivotal/plugin.py#L97)):*
> ```Python
>             story.add_note(comment)
>         except pyvotal.exceptions.AccessDenied:
>             raise forms.ValidationError(_('Access denied through Pivotal API.'))
> ```

you probably don't want to use the `forms.ValidationError` here (or elsewhere in the plugin) since we're not using django forms anymore. Instead you can use `sentry.plugins.bases.issue2.PluginError` and the `Issue2` plugin class should handle displaying the message to the user correctly (see https://github.com/getsentry/sentry/blob/master/src/sentry/plugins/bases/issue2.py#L289 -- and you can look at the github plugin for examples too)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5)*

macqueen:
Reviewed 3 of 5 files at r1, 1 of 2 files at r2.
Review status: 4 of 5 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5)*

macqueen:
Review status: 4 of 5 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/sentry_plugins/pivotal/plugin.py, line 54 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSn0jCP0vwOo_mV-Vl:-KQSn0jCP0vwOo_mV-Vm:biyynrn) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L54)):*
> ```Python
>             'type': 'select',
>             'has_autocomplete': True,
>             'help': ('Search Pivotal Stories by name or description.')
> ```

you don't need parens here

---

*[src/sentry_plugins/pivotal/plugin.py, line 72 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSqD0z0qsGjMUpX_lC:-KQSqD0z0qsGjMUpX_lD:b-an8wcd) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L72)):*
> ```Python
>         client = self.get_api_client(group.project)
> 
>         try:
> ```

For this try/except block, I'd suggest doing something like this:

```python

try:
    project = client.projects.get(self.get_option('project', group.project))
    stories = project.stories.all()
except (pyvotal.exceptions.PyvotalException, pyvotal.exceptions.AccessDenied) as error:
    status = 400 if isinstance(error, PluginError) else 502
    return Response({'error_type': 'validation',
                     'errors': [{'__all__': GENERIC_ERROR}],
                     }, status=status)

issues = []
for story in stories:
    if query.lower() in story.name.lower() or query.lower() in story.description.lower():
        issues.append({'text': '(#%s) %s)' % (story.id, story.name),
                       'id': story.id})
return Response({field: issues})

```
Reasoning:
- Those error classes based on https://github.com/fullboar/pyvotal/blob/master/pyvotal/exceptions.py -- if it turns out that the library sometimes raises other exceptions, we can add them as they happen.
- You want an exception to actually be thrown if, say, `story.name` is None and it won't be if you keep your for loop in the try block and are catching all `Exception`.

---

*[src/sentry_plugins/pivotal/plugin.py, line 77 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSnikkGtPev0fx1ArO:-KQSniklER0vFQ7j1Air:bmik83o) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L77)):*
> ```Python
>             issues = []
>             for story in stories:
>                 if query.lower() in story.name.lower() or query.lower() in story.description.lower():
> ```

since you're doing `query.lower()` multiple times and in a loop, i'd do something like `query = query.lower()` before the for loop

---

*[src/sentry_plugins/pivotal/plugin.py, line 97 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSvUKFhOT0kmJUAPzw:-KQSvUKFhOT0kmJUAPzx:bg4x6b0) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L97)):*
> ```Python
>             story = project.stories.get(form_data['issue_id'])
>             story.add_note(comment)
>         except:
> ```

here it'd probably also be good to limit to exception classes we think the library may throw

---

*[src/sentry_plugins/pivotal/plugin.py, line 108 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSx5dJ3CaXZp5PGcuX:-KQSx5dKSGsIguR5yyHQ:bhbhjj1) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L108)):*
> ```Python
>             story.name = force_text(form_data['title'], errors='replace')
>             story.description = force_text(form_data['description'], errors='replace')
>             story.labels = "sentry"
> ```

here too, i'd try to limit what you put in the try block. does `client.Story()` actually make a request to the pivotal api? Because if not, I'd move this line and above to before the try block

---

*[src/sentry_plugins/pivotal/plugin.py, line 111 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQSvdbI_t-5fOa8Yw4x:-KQSvdbI_t-5fOa8Yw4y:b1s273b) ([raw file](https://github.com/getsentry/sentry-plugins/blob/abf6cb75f0e8e060042d6234b681fa0c490d20d2/src/sentry_plugins/pivotal/plugin.py#L111)):*
> ```Python
>             project = client.projects.get(self.get_option('project', group.project))
>             story = project.stories.add(story)
>         except:
> ```

same here about exception classes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5)*

mattrobenolt:
Reviewed 2 of 5 files at r1.
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[setup.py, line 21 \[r4\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQWkFe1JPZ7h1H_XsYv:-KQWkFe1JPZ7h1H_XsYw:b-ofddm9) ([raw file](https://github.com/getsentry/sentry-plugins/blob/e8e36ddb616479f57d6a13c3f8b68f0d938fa5a5/setup.py#L21)):*
> ```Python
> ]
> 
> install_requires = ['pyvotal>=0.2.1']
> ```

To be safe, this should also be bounded on the upper side, like `pyvotal>=0.2.1,<0.3.0` so they don't introduce an API breaking change. It's minor and no guarantee, but it's best practice to at least make an attempt, but I won't block because of that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5)*

macqueen:
:lgtm: -- can you squash your commits before merging?

---

Reviewed 1 of 1 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/sentry_plugins/pivotal/plugin.py, line 121 \[r5\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQWs-e5E7PU5CZ8mhLx:-KQWs-e5E7PU5CZ8mhLy:b-krcd0q) ([raw file](https://github.com/getsentry/sentry-plugins/blob/6e07d87cdcbb293ae87bb12f4141d0a31870541e/src/sentry_plugins/pivotal/plugin.py#L121)):*
> ```Python
> 
>     def get_issue_url(self, group, issue_id, **kwargs):
>         return 'https://www.pivotaltracker.com/story/show/%s' % (issue_id)
> ```

these parens aren't needed, but doesn't need to block this

---

*[src/sentry_plugins/pivotal/plugin.py, line 137 \[r5\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-KQWqhrCO76c_ZOVDjN-:-KQWqhrCO76c_ZOVDjN0:b76o73d) ([raw file](https://github.com/getsentry/sentry-plugins/blob/6e07d87cdcbb293ae87bb12f4141d0a31870541e/src/sentry_plugins/pivotal/plugin.py#L137)):*
> ```Python
>             'type': 'text',
>             'placeholder': 'e.g. a9877d72b6d13b23410a7109b35e88bc',
>             'help': 'Enter your API Token (found on %s).' % ('<a href="https://www.pivotaltracker.com/profile">pivotaltracker.com/profile</a>')},
> ```

you don't need to use string formatting here since it isn't a variable, but also not blocking

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5#-:-KQWsZxulttQt4Fjl2Q3:b-9pgfvr)*

macqueen:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/5)*

PR #6, opened 2016-08-30
Add Sentry lint utilities

dcramer:
@macqueen

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/6)


hooks/pre-commit
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
dcramer: tbh we should probably work towards making sentry's pre-commit hook reusable (since the VIRTUAL_ENV check here is pretty useful, and its copypasta)

macqueen:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/6)*

PR #7, opened 2016-08-31
Add HipChat (Connect) plugin

dcramer:
- Works with both legacy UI as well as upcoming React changes (getsentry/sentry#4058)
- Removes use of SENTRY_URL_PREFIX, which is invalid, but happens to work on sentry.io

@macqueen

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/7)


dcramer:
This still needs to manage asset loading. Right now it can generate the .js file, but theres nothing to do it automatically (e.g. as part of setup.py install), and there's nothing to actually load the .js in Sentry itself.

dcramer:
Of note, I'm going to test this against master, and if it works we can probably get the core of this in, and do a follow up PR which finishes the React transition (as it won't even be used in master right now).

dcramer:
Seems to still work fine without the React changes

dcramer:
Also I fixed like 10 ``</p>`` tags that @mitsuhiko forgot

dcramer:
FYI the only changes here are:

- Remove SENTRY_URL_PREFIX (which adds the regexp cache, makes some things dynamic)
- Add the JS prototype (unused right now)

macqueen:
Reviewed 31 of 38 files at r1, 1 of 2 files at r2, 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[webpack.config.js, line 50 \[r1\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/7#-KQaykSttfDS7lQ44lUt:-KQaykSttfDS7lQ44lUu:b-p4mnge) ([raw file](https://github.com/getsentry/sentry-plugins/blob/97138d32d169501921ef45084eac8b76a7479648/webpack.config.js#L50)):*
> ```JavaScript
>       path: path.join(__dirname, distPath),
>       // libraryTarget: 'var',
>       // library: ['Sentry', 'plugins', app],
> ```

any reason to not delete these?

---

*[src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx, line 65 \[r3\]](https://reviewable.io:443/reviews/getsentry/sentry-plugins/7#-KQcYuq27vClc9o6LrGn:-KQcYuq3glp0io9mGBl8:b-1e32qq) ([raw file](https://github.com/getsentry/sentry-plugins/blob/421a777bf7aacd1ff3ba1014065a5a997eeafe31/src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx#L65)):*
> ```JSX
> 
>     let tenants = this.state.tenants;
>     if (tenants.length === 0)
> ```

i'd just do `if (!tenants.length)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/getsentry/sentry-plugins/7)*

PR #8, opened 2016-09-02
Add validation for GitHub repo

dcramer:
Requires getsentry/sentry#4058

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/8)


dcramer:
This currently will have issues as its not guaranteed the user has connected GitHub at this stage. Id like to do this kind of change to improve UX, but we'd first need to say "require the user connect to github on plugin enable".

PR #11, opened 2016-09-05
Add GitLab plugin

dcramer:
Requires getsentry/sentry#4081

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/11)


dcramer:
Actually this needs to handle the "link existing" bit yet

dcramer:
This is done -- can't reasonably handle the link flow due to limitations of GitLab's API.

src/sentry_plugins/gitlab/endpoints.py
```diff
@@ -0,0 +1,25 @@
...
+#         repo = self.get_option('repo', group.project)
+#         client = self.plugin.get_client(group.project)
+#         response = client.search_issues(repo, query)
+
+#         issues = [{
+#             'text': '(#%s) %s' % (i['number'], i['title']),
+#             'id': i['number']
+#         } for i in response.get('items', [])]
+
+#         return Response({field: issues})
```
macqueen: sad :(

dcramer: I was going to leave it as I *think* GitLab is going to get in some APIs we need, and it's a good example bit we can pull from moving forward.

PR #16, opened 2016-09-12
Refactor to use latest Pivotal API version

kjlundsgaard:
The python client (pyvotal) that we were using for the Pivotal API was using an outdated endpoint, so I refactored. This should also fix https://sentry.io/sentry/sentry/issues/158275445/
@getsentry/plugins

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/16)


src/sentry_plugins/pivotal/plugin.py
```diff
@@ -62,52 +62,87 @@ def view_autocomplete(self, request, group, **kwargs):
...
         if field != 'issue_id' or not query:
             return Response({'issue_id': []})
-        client = self.get_api_client(group.project)
+        query = query.encode('utf-8')
+        _url = '%s?%s' % (self.build_api_url(group, 'search'), urlencode({'query': query}))
+        try:
+            req = self.make_api_request(group.project, _url)
+            body = safe_urlread(req)
+        except (requests.RequestException, PluginError) as e:
+            return self.handle_api_error(e)
```
macqueen: do you actually define a `handle_api_error` method?

src/sentry_plugins/pivotal/plugin.py
```diff
@@ -62,52 +62,87 @@ def view_autocomplete(self, request, group, **kwargs):
...
+        } for i in json_resp['stories']['stories']]
+
         return Response({field: issues})
 
     def link_issue(self, request, group, form_data, **kwargs):
         comment = form_data.get('comment')
         if not comment:
             return
-        client = self.get_api_client(group.project)
+        print form_data['issue_id']
```
macqueen: you can prob delete this :)

src/sentry_plugins/pivotal/plugin.py
kjlundsgaard: Added

src/sentry_plugins/pivotal/plugin.py
kjlundsgaard: Done

PR #20, opened 2016-09-19
Fix auth redirect behavior

dcramer:
Only bind authentication details when we send the user to the login form.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/20)


dcramer:
This relies on changes in https://github.com/getsentry/sentry/pull/4165

PR #22, opened 2016-09-20
[github] Improve error handling

dcramer:
- Add GitHub client abstraction
- Handle 404 and unauthorized errors better

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/22)


dcramer:
screenshot 2016-09-20 10 52 55

src/sentry_plugins/github/plugin.py
```diff
@@ -70,101 +73,84 @@ def get_link_existing_issue_fields(self, request, group, event, **kwargs):
...
-            req = self.make_api_request(request.user, _url)
-            body = safe_urlread(req)
-        except (requests.RequestException, PluginError) as e:
-            raise PluginError(u'Error communicating with GitHub: %s' % e)
+    def message_from_error(self, exc):
+        if isinstance(exc, ApiUnauthorized):
+            return ERR_UNAUTHORIZED
+        elif isinstance(exc, ApiError):
+            if exc.code == 404:
+                return ('GitHub returned a 404 Not Found error. If this repository exists, ensure you have access to it and that Sentry is being allowed to read data from this organization (Account Settings > Authorized Applications > Sentry).')
```
macqueen: why is the linter not complaining about how long this line is?

dcramer: I think we the line length checks disabled since the hard rule of X chars generally isn't good. That said this is crazy long so I fixed.

PR #24, opened 2016-09-20
Tenant.auth_user may be NULL

mattrobenolt:
Fixes SENTRY-1Z9

@macqueen @dcramer I'm not sure if whatever handles this response will account for this, but the column is definitely nullable.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/24)


dcramer:
fyi i think we have tests for this endpoint so could probably extend it to have one without auth user

mattrobenolt:
Added test and added fix for the UI to handle this.

PR #25, opened 2016-09-20
Auto hook up webpack watcher process for `devserver`

mattrobenolt:
@macqueen @dcramer

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/25)

### DIFF
--- a/src/sentry_plugins/__init__.py
+++ b/src/sentry_plugins/__init__.py
@@ -3,5 +3,23 @@ from __future__ import absolute_import
 try:
     VERSION = __import__('pkg_resources') \
         .get_distribution('sentry-plugins').version
-except Exception, e:
+except Exception as e:
     VERSION = 'unknown'
+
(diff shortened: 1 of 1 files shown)


mattrobenolt:
Ok, I got this thing running locally, and confirmed that this does work.

src/sentry_plugins/__init__.py
```diff
@@ -3,5 +3,23 @@
 try:
     VERSION = __import__('pkg_resources') \
         .get_distribution('sentry-plugins').version
-except Exception, e:
+except Exception as e:
     VERSION = 'unknown'
+
+# Try to hook our webhook watcher into the rest of the watchers
+# iff this module is installed in editable mode.
```
macqueen: iff

mattrobenolt: Why does nobody know what this means? 😡 

https://en.wikipedia.org/wiki/If_and_only_if

PR #27, opened 2016-09-21
[github] improve link issue flow

dcramer:
This moves the title fetching logic into link_issue, and also ensures that failing to fetch an issue will raise a PluginError (rather than returning a Response object).

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-plugins/27)


dcramer:
Requires https://github.com/getsentry/sentry/pull/4194

PR #29, opened 2016-09-25
[github] Automatically store commit metadata from webhooks

dcramer:
This wires up a PushEvent webhook which will create Commit objects in Sentry. Repositories are keyed on a github prefix, and the existing name is used.


dcramer:
Blocking on https://github.com/getsentry/sentry/pull/4209

dcramer:
Remaining TODO:

[ ] Create webhook on enable
[ ] Remove webhook on disable
[ ] Mechanism to rotate secret (and auto create)

dcramer:
Main complexity now will be what happens on enable/disable. We can't wipe commits as they might be created elsewhere, but we may end up with a huge pollution of data if you're switching between repos, though arguably you shouldn't do that.

I think a good step forward would be to let you add multiple repositories (somehow). Maybe by linking GitHub that allows you to add a GitHub repository elsewhere? We could then bind commits to repos.

dcramer:
More brain dump.

We could add a repositories section to the organization. We know that eventually when we let you plugin a repo we'll need it anyways, and then commits get associated with that repository. As part of that, the GitHub plugin could allow you to add repositories to that list (some kind of simple extension point?).

dcramer:
This works barring automatic enable/disable. We could land the initial bits of this and wire it up manually to the repos we need to prototype.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -0,0 +1,169 @@
...
+            name='github:{}'.format(event['repository']['full_name']),
+        )[0]
+
+        for commit in event['commits']:
+            if not commit['distinct']:
+                continue
+
+            author_email = commit['author']['email']
+            if '@' not in author_email:
+                author_email = '{}@localhost'.format(
```
mattrobenolt: Need a `u` here or we'll get a `UnicodeEncodeError` error.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class PushEventWebhook(Webhook):
+    # https://developer.github.com/v3/activity/events/types/#pushevent
+    def __call__(self, organization, event):
+        authors = {}
+
+        repo = Repository.objects.get_or_create(
+            organization_id=organization.id,
+            name='github:{}'.format(event['repository']['full_name']),
```
mattrobenolt: Probably need `u` here too. I assume a repository name may have unicode characters.

dcramer:
@macqueen shipping this and gonna manually wire up webhooks for Sentry

PR #33, opened 2016-09-28
Add tests to assert that all entry_points are registered

mattrobenolt:
@getsentry/plugins


tests/jira/test_plugin.py
```diff
@@ -0,0 +1,34 @@
...
+        return JiraACPlugin()
+
+    @fixture
+    def request(self):
+        return RequestFactory()
+
+    def test_conf_key(self):
+        assert self.plugin.conf_key == 'jira-ac'
+
+    def assertEntryPointInstalled(self, plugin, name):
```
mattrobenolt: I'm going to move this into `sentry.testutils:TestCase` and use it for all tests assumign this works.

mattrobenolt: https://github.com/getsentry/sentry/pull/4236

PR #35, opened 2016-09-30
[github] utilize InvalidIdentity control flow

dcramer:
Requires https://github.com/getsentry/sentry/pull/4249


src/sentry_plugins/github/plugin.py
```diff
@@ -114,6 +108,8 @@ def message_from_error(self, exc):
     def raise_error(self, exc):
         if not isinstance(exc, ApiError):
             self.logger.exception(six.text_type(exc))
+        if isinstance(exc, ApiUnauthorized):
```
dcramer: this is technically not 100% correct, they might simply not have access to the repo

PR #38, opened 2016-10-05
move jira to react

macqueen:
still need to write tests @getsentry/plugins @getsentry/product 

configuration in read only mode (will fix disabled styles for these in sentry repo)
screenshot 2016-10-04 18 03 50

configuration in edit mode:
1.
screenshot 2016-10-04 18 04 43
2.
screenshot 2016-10-04 18 04 56
3.
screenshot 2016-10-04 18 05 20

issue creation
screenshot 2016-10-04 18 07 56


dcramer:
omg2

macqueen:
did some more qa and fixed auto creation. think this is pretty good to go.

src/sentry_plugins/jira/client.py
```diff
@@ -0,0 +1,181 @@
...
+from simplejson.decoder import JSONDecodeError
+from BeautifulSoup import BeautifulStoneSoup
+from django.utils.datastructures import SortedDict
+
+log = logging.getLogger(__name__)
+
+CACHE_KEY = "SENTRY-JIRA-%s-%s"
+
+
+class JIRAError(Exception):
```
dcramer: we could inherit from ApiError here

src/sentry_plugins/jira/plugin.py
```diff
@@ -0,0 +1,496 @@
...
+        # issuetype assigned.
+        if not form_data.get('issuetype'):
+            raise PluginError('Issue Type is required.')
+
+        jira_project_key = self.get_option('default_project', group.project)
+        client = self.get_jira_client(group.project)
+        meta = client.get_create_meta_for_project(jira_project_key)
+
+        if not meta:
+            raise PluginError('Something went wrong. Check your plugin configuration.')
```
dcramer: follow: could we do something better here for what the error might actually be?

src/sentry_plugins/jira/static/jira/components/issueActions.jsx
```diff
@@ -0,0 +1,95 @@
...
+            })}
+          </div>
+        );
+        let required = field.required != null ? field.required : true;
+        if (required) {
+          requiredFields.push(el);
+        } else {
+          optionalFields.push(el);
+        }
+      });
```
benvinegar: These days, I'd probably implement this whole section like this:

- Create a list of all fields
- From that list, filter it into two lists (required + optional)

Something like:

```javascript
let fields = this.state.createFieldList.map(field => {
  return <div key={field.name}>/*...*/</div>;
});
let isRequired = (f) => f.required !== null ? field.required : true;
let requiredFields = fields.filter(field => isRequired(f));
let optionalFields = fields.filter(field => !isRequired(f));
```

This is a little more functional, and while it adds a little redundancy (iterates through the list twice), I feel it's cleaner to read (`map` and `filter` signal intent, whereas `forEach` can be anything).

src/sentry_plugins/jira/static/jira/components/settings.jsx
```diff
@@ -0,0 +1,149 @@
...
+      if (this.state.page !== 2) {
+        this.setState({page: this.state.page + 1});
+      }
+      this.onSaveSuccess(this.onSaveComplete);
+      return;
+    }
+    this.api.request(this.getPluginEndpoint(), {
+      data: this.state.formData,
+      method: 'PUT',
+      success: this.onSaveSuccess.bind(this, data => {
```
benvinegar: I miss auto-binding ...

src/sentry_plugins/jira/static/jira/components/issueActions.jsx
benvinegar: I mean, I think that looks pretty good too, but you don't have to change anything – this was mostly just a conversation starter :)

PR #39, opened 2016-10-07
[slack] initial import

dcramer:


tests/slack/test_plugin.py
```diff
@@ -0,0 +1,63 @@
...
+from sentry_plugins.slack.plugin import SlackPlugin
+
+
+class SlackPluginTest(TestCase):
+    @fixture
+    def plugin(self):
+        return SlackPlugin()
+
+    def test_conf_key(self):
+        assert self.plugin.conf_key == 'slack'
```
mattrobenolt: Should just be:

```python
def test_entry_point(self):
    self.assertAppInstalled('slack', 'sentry_plugins.slack')
    self.assertPluginInstalled('slack', self.plugin)
```

And subclass `PluginTestCase` instead of `TestCase`.

PR #40, opened 2016-10-07
Hipchat improvements

kjlundsgaard:
Makes configuration flow better and fixes a couple bugs
@getsentry/plugins


src/sentry_plugins/hipchat_ac/plugin.py
```diff
@@ -117,27 +115,28 @@ def is_configured(self, project):
...
-        return render_to_string('sentry_hipchat_ac/configure_plugin.html', dict(
-            plugin=self,
-            plugin_test_results=test_results,
-            on_premise=is_on_premise(),
-            tenants=list(project.hipchat_tenant_set.select_related('auth_user')),
-            descriptor=self.get_descriptor(),
-            install_url=self.get_install_url(),
-            context_instance=RequestContext(request)
-        ))
+    # def configure(self, request, project=None):
```
dcramer: id just kill this now rather than commenting it out

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
```diff
@@ -58,14 +58,45 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
...
   }
-
-  renderTenants() {
+  renderLink(url, metadata) {
+    let tenants = this.state.tenants;
+    if (!tenants || !tenants.length) {
+      return (
+        <div>
+          <p>
+            Hipchat notifications are managed through Hipchat itself.  After you
```
dcramer: "HipChat" vs "Hipchat" as far as I understand

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
```diff
@@ -58,14 +58,45 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
...
+            added the Sentry integration directly into Hipchat you can add projects
```
dcramer: After you've added

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
```diff
@@ -3,19 +3,36 @@
   {% if tenant.auth_user %}
     <h3>Authorization</h3>
     <p>
-      This room is authenticated as <strong>{{ tenant.auth_user.username }}</strong> and
-      was authorized to have access to the following organizations:
+      You are currently signed in as <strong>{{ current_user.username }}</strong>.
+      <a href="{% url 'sentry-hipchat-ac-sign-out' %}?signed_request={{ context.signed_request }}">Change user?</a>
+    <p>
```
dcramer: omg cant handle no ``</p>`` tags what have you created @mitsuhiko

src/sentry_plugins/hipchat_ac/plugin.py
kjlundsgaard: done

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
kjlundsgaard: done

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
kjlundsgaard: done

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
kjlundsgaard: lol

src/sentry_plugins/hipchat_ac/views.py
```diff
@@ -369,6 +397,7 @@ def new_f(request, *args, **kwargs):
         # put something here so that sentry does not overwrite it
         # with deny.
         resp['X-Frame-Options'] = 'allow'
+        resp['Content-Security-Policy'] = 'frame-ancestors https://sentry.hipchat.com/'
```
dcramer: /cc @mattrobenolt

mattrobenolt: We shouldn't need to also add this header here. And to be fair, for `getsentry`, this will not actually work and will be clobbered anyways. The above `X-Frame-Options` rule should be sufficient.

Also, this is a hardcoded value that would only work for us, so this pattern should be avoided and we'd want to leverage a configuration setting to define what this would be.

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
```diff
@@ -58,14 +58,45 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
...
   }
-
-  renderTenants() {
+  renderLink(url, metadata) {
+    let tenants = this.state.tenants;
+    if (!tenants || !tenants.length) {
+      return (
+        <div>
+          <p>
+            HipChat notifications are managed through HipChat itself.  After you've
```
dcramer: @cameronmcefee might have some thoughts on language, but you use directly twice :)

src/sentry_plugins/hipchat_ac/views.py
kjlundsgaard: 👍  removed

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
kjlundsgaard: Oh this is all language that was already there, but I'm happy to change

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
```diff
@@ -58,20 +58,53 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
       }
     });
   }
+  renderLink(url, metadata) {
+    let tenants = this.state.tenants;
+    if (!tenants || !tenants.length) {
+      if(metadata.onPremise) {
```
macqueen: i think we usually do space between `if` and `(`

src/sentry_plugins/hipchat_ac/static/hipchat_ac/components/settings.jsx
```diff
@@ -58,20 +58,53 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
...
+            </p>
+            <p>
+              <a href={url}
+                 className="btn btn-primary"
+                 target="_blank">Enable Integration</a>
+            </p>
+          </div>
+        );
+      }
+      else {
```
macqueen: also usually would bump this up to previous line so it's `} else {`

src/sentry_plugins/hipchat_ac/views.py
```diff
@@ -235,21 +253,33 @@ def post(self, request):
...
+        #     pass
+        # else:
+        #     old_tenant.delete()
+
+        # tenant = Tenant.objects.create(
+        #     id=data['oauthId'],
+        #     room_id=room_id,
+        #     secret=data['oauthSecret'],
+        #     capdoc=capdoc,
+        # )
```
macqueen: can the above commented out code just be deleted?

PR #44, opened 2016-10-10
[github] fix push event

dcramer:
The docs were seemingly a lie. Looking at GitHub I see the header sent showing 'push' instead of 'PushEvent'.


dcramer:
Looks like form the time I implemented this to the time we deployed it the docs were updated :(

PR #61, opened 2016-10-20
Open Sentry login in popup and reload configuration page after login

kjlundsgaard:
Opens a new window to for user to log in and receives a message from the login window after. Relies on https://github.com/getsentry/sentry/pull/4398

@getsentry/plugins @benvinegar


src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
```diff
@@ -1,5 +1,38 @@
...
+    var left = screen.width / 2 - 300
+    popup = window.open(
+      (extractOrigin() + "/auth/login?next=" + encodeURIComponent('/auth/close/')),
+      "sentryLogin",
+      "modal, resizable, scrollbars, height=700, width=600, top="+top+", left="+left
+    );
+
+    window.addEventListener('message', function (evt) {
+      if (event.origin !== extractOrigin()) {
+        console.log("wrong domain");
```
dcramer: maybe throw a hard error here, so if Sentry is bound itll automatically handle it.

e.g.

```
throw new Error('message event received from invalid origin: ' + event.origin);
```

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
```diff
@@ -1,5 +1,38 @@
...
+    return origin;
+  }
+
+  var popup;
+
+  function openLogin() {
+    var top = screen.height / 2 - 350
+    var left = screen.width / 2 - 300
+    popup = window.open(
+      (extractOrigin() + "/auth/login?next=" + encodeURIComponent('/auth/close/')),
```
mattrobenolt: Why do we need to do this `extractOrigin()`? Can't we use our `{% absolute_uri %}` template tag and just spit it out correctly?

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
```diff
@@ -1,5 +1,38 @@
...
+    var origin = protocol + '//' + host;
+    return origin;
+  }
+
+  var popup;
+
+  function openLogin() {
+    var top = screen.height / 2 - 350
+    var left = screen.width / 2 - 300
+    popup = window.open(
```
kjlundsgaard: Initially I thought I might be sending a message with postMessage, but it does not need to be bound. I'll make this change.

mattrobenolt:
With limited context, I have to ask, "don't browsers just by default block this now?" I don't have the context on why we need this, or if there's an alternative, but I think all `window.open()` is just basically rejected by default now without prompting the user to allow it.

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
kjlundsgaard: Didn't realize you could do that. I'll update this on the sentry side as well.

benvinegar:
@kjlundsgaard – can you update this now that @mattrobenolt landed getsentry/sentry#4410?

src/sentry_plugins/hipchat_ac/templates/sentry_hipchat_ac/configure.html
```diff
@@ -1,5 +1,32 @@
...
 {% block body %}
+
+<script>
+// open sentry login in another window and then receive a message of whether user is logged in. If user is logged in, refresh the page, so that user can continue through configuration flow
+
+  function openLogin() {
+    var top = screen.height / 2 - 350
+    var left = screen.width / 2 - 300
+    window.open(
+      ('{% system_origin %}' + "/auth/login?next=" + encodeURIComponent('/auth/close/')),
```
mattrobenolt: _Technically_ no reason to need to append here anymore, but doesn't particularly matter. `'{% system_origin %}/auth/login?next='` would be just fine.

PR #66, opened 2016-10-26
[jira] fix invalid issue type being saved for projects

macqueen:
in some cases we were using an issue type from the wrong project in jira. fixes https://github.com/getsentry/sentry-plugins/issues/65

@getsentry/plugins


src/sentry_plugins/jira/plugin.py
```diff
@@ -156,6 +156,13 @@ def get_new_issue_fields(self, request, group, event, **kwargs):
             issue_type = self.get_option('default_issue_type', group.project)
 
         issue_type_meta = self.get_issue_type_meta(issue_type, meta)
+        issue_type_choices = self.make_choices(meta['issuetypes'])
```
macqueen: these changes make sure people who have invalid saved values won't have a problem

src/sentry_plugins/jira/static/jira/components/settings.jsx
```diff
@@ -69,8 +69,14 @@ class Settings extends plugins.BasePlugin.DefaultSettings {
       this.onSaveSuccess(this.onSaveComplete);
       return;
     }
+    let formData = Object.assign({}, this.state.formData);
+    // if the project has changed, it's likely these values aren't valid anymore
+    if (formData.default_project !== this.state.initialData.default_project) {
```
macqueen: and these should prevent it from happening at all for new configurations

PR #67, opened 2016-10-26
allow redirects for github api

macqueen:
fixes https://sentry.io/sentry/sentry/issues/170917210/

i think this just got reintroduced: https://github.com/getsentry/sentry-github/pull/33

@getsentry/plugins


dcramer:
damn wish we had history on Sentry now :(

PR #68, opened 2016-11-01
jira: generate a more reliable cache key

mattrobenolt:
Using the url itself opens itself up to a lot of failures. Either the
key being too long or containing characters that are invalid.

Fixes SENTRY-293 and I'm sure a few others.


dcramer:
also https://github.com/getsentry/sentry/pull/4460

PR #69, opened 2016-11-02
add asana to issue tracking integrations

macqueen:
still need to write tests, but otherwise this is ready

https://github.com/getsentry/sentry/pull/4452 needs to be merged first

@getsentry/plugins


tests/asana/test_plugin.py
```diff
@@ -0,0 +1,132 @@
...
+            provider=self.plugin.auth_provider,
+            extra_data={'access_token': 'foo'},
+        )
+
+        assert self.plugin.link_issue(request, group, form_data) == {
+            'title': 'Hello',
+        }
+        request = responses.calls[-1].request
+        payload = json.loads(request.body)
+        print payload
```
mattrobenolt: `print`

src/sentry_plugins/asana/plugin.py
```diff
@@ -0,0 +1,251 @@
...
+                view_method_name='view_autocomplete',
+                plugin=self,
+            )),
+        ]
+
+    def is_configured(self, request, project, **kwargs):
+        return bool(self.get_option('workspace', project))
+
+    def has_workspace_access(self, workspace, choices):
+        return workspace in set([c[0] for c in choices])
```
mattrobenolt: Pro Python tip: Set's have a shorter syntax here.

```python
{c[0] for c in choices}
```

But alternatively, you're creating a set, just to check if something is contained in it. If the goal is to be more efficient, you can write it as:

```python
for c, _ in choices:
  if workspace == c:
    return True
return False
```

PR #73, opened 2016-11-02
fix handling of number fields in jira plugin

macqueen:
@getsentry/plugins


macqueen:
fixes https://github.com/getsentry/sentry-jira/issues/108 in this repo

src/sentry_plugins/jira/plugin.py
```diff
@@ -395,6 +395,11 @@ def create_issue(self, request, group, form_data, **kwargs):
                         v = [v]
                     elif schema.get('custom') == JIRA_CUSTOM_FIELD_TYPES.get('textarea'):
                         v = v
+                    elif schema['type'] == 'number':
+                        try:
+                            v = int(v)
+                        except ValueError:
```
mattrobenolt: This could also raise a `TypeError` if `v` were `None` or some other data structure entirely, so unless you're sure, I'd do `except (TypeError, ValueError):`

src/sentry_plugins/jira/plugin.py
```diff
@@ -395,6 +395,11 @@ def create_issue(self, request, group, form_data, **kwargs):
                         v = [v]
                     elif schema.get('custom') == JIRA_CUSTOM_FIELD_TYPES.get('textarea'):
                         v = v
+                    elif schema['type'] == 'number':
+                        try:
+                            v = int(v)
```
mattrobenolt: Just double check that this `number` value won't work with floats. If it does, this will truncate them.

src/sentry_plugins/jira/plugin.py
macqueen: there's a check above for `if v:`

src/sentry_plugins/jira/plugin.py
macqueen: 👍

PR #74, opened 2016-11-02
actually pass issue type as request param when user changes issue type

macqueen:
different issue types have different fields, so we need to regenerate the form when the user changes the issue type

@getsentry/plugins


src/sentry_plugins/jira/static/jira/components/issueActions.jsx
```diff
@@ -15,7 +15,7 @@ class IssueActions extends plugins.DefaultIssuePlugin.DefaultIssueActions {
     if (name === 'issuetype') {
       state.state = FormState.LOADING;
       this.setState(state, this.onLoad.bind(this, () => {
-        this.api.request(this.getPluginCreateEndpoint(), {
+        this.api.request(this.getPluginCreateEndpoint() + '?issue_type=' + value, {
```
mattrobenolt: Wouldn't you need to uri encode `value`?

mattrobenolt: From my little understnaidng of this code, there's nothing that'd prevent `value` from being like, `&foo=1` or something else that'd break the url. So psure this would need to be uri encoded first.

PR #81, opened 2016-11-04
fix KeyError u'\u015b' in jira_ac plugin

macqueen:
fixes https://sentry.io/sentry/sentry/issues/177001759/


src/sentry_plugins/jira_ac/utils.py
```diff
@@ -13,7 +13,7 @@ class JIRAError(Exception):
 
 def percent_encode(val):
     # see https://en.wikipedia.org/wiki/Percent-encoding
-    return quote(val).replace('%7E', '~').replace('/', '%2F')
+    return quote(val.encode('utf8')).replace('%7E', '~').replace('/', '%2F')
```
mattrobenolt: Do `.encode('utf8', replace='errors')` to prevent it from erroring if it's not utf8 compatible.

Issue #79, opened 2016-11-04
Make secret fields indicate that a value exists already more obviously


mattrobenolt:
GitHub does this:

![image](https://cloud.githubusercontent.com/assets/375744/20021470/225d39d8-a293-11e6-973c-09b17e524804.png)
![image](https://cloud.githubusercontent.com/assets/375744/20021474/26de366a-a293-11e6-8228-0e860a969d83.png)

PR #83, opened 2016-11-09
make it more obvious when we have a saved value for integrations

macqueen:
@getsentry/plugins 

related to https://github.com/getsentry/sentry/pull/4492
fixes https://github.com/getsentry/sentry-plugins/issues/79


macqueen:
tests will stop failing once i merge the sentry changes

PR #85, opened 2016-11-09
hipchat-ac: validate required scopes for forms and confirmation

mattrobenolt:
Without this check, it was possible to work around and add a Hipchat
plugin when you didn't have permission to do so.

This also cleans up the UX a bit to only show and present organizations/teams
with the correct permission in the first place.

Needs: https://github.com/getsentry/sentry/pull/4499


mattrobenolt:
This also affects jira_ac, but we should do that as a separate PR since it's a different fix there.

PR #86, opened 2016-11-14
attempt to fix account custom field in jira

macqueen:
@getsentry/plugins


macqueen:
@mattrobenolt yeah other people have complained about this issue (https://jira.atlassian.com/browse/JRA-45371). since this is a relatively easy change, i figured it was worth doing, but in general, i think we're not really going to be able to support custom fields very well.

PR #93, opened 2016-11-26
[segment] initial working code

dcramer:


src/sentry_plugins/segment/plugin.py
```diff
@@ -0,0 +1,62 @@
...
+        rl_key = 'segment:{}'.format(md5_text(write_key).hexdigest())
+        # limit segment to 50 requests/second
+        if ratelimiter.is_limited(rl_key, limit=50, window=1):
+            return
+
+        payload = {
+            'userId': user_id,
+            'event': 'Error Captured',
+            'properties': {
+                'eventId': event.event_id,
```
dcramer: some additional things we might want to add here:

- transaction name
- http data (if available): url, method?
- project name?

src/sentry_plugins/segment/plugin.py
```diff
@@ -0,0 +1,62 @@
...
+            return
+
+        rl_key = 'segment:{}'.format(md5_text(write_key).hexdigest())
+        # limit segment to 50 requests/second
+        if ratelimiter.is_limited(rl_key, limit=50, window=1):
+            return
+
+        payload = {
+            'userId': user_id,
+            'event': 'Error Captured',
```
dcramer: this is not an upstream event -- the only thing applicable today is "Application Crashed" but it's my opinion we shouldn't send that as its not correct and it might be duplicate data if they're already using the client-side Segment integration

sperand-io: @dcramer agreed — this event name makes sense to me! we'll pull this into the spec

src/sentry_plugins/segment/plugin.py
sperand-io: on this, i defer entirely to you all on what kind of contextual data could help our mutual customers. those all make sense to me.

dcramer:
@sperand-io we generally avoid deps as they often have other requirements and have caused issues in the past. I think if we were doing something more complex than constructing a simple HTTP payload we'd use it.

Aside, our internal HTTP stuff enforces some guarantees, which is a secondary reason. They dont matter much here, but they're things like preventing queries to internal services (e.g. if your DNS was hacked), and setting user agents.

I'm going to give more thought to the data that we make available here. There's some complexity in that Sentry supports a variety of event types, and maybe we would want to spec out each one. e.g. an exception event, a csp report event, a log message event.

src/sentry_plugins/segment/plugin.py
```diff
@@ -0,0 +1,94 @@
...
+            'release': event.get_tag('sentry:release'),
+        }
+        if 'sentry.interfaces.Http' in event.interfaces:
+            http = event.interfaces['sentry.interfaces.Http']
+            headers = http.headers
+            if not isinstance(headers, dict):
+                headers = dict(headers or ())
+
+            props.update({
+                'requestUrl': http.url,
```
dcramer: @sperand-io are you already doing any kind of standard naming on these HTTP properties?

dcramer:
requires https://github.com/getsentry/sentry/pull/4685

PR #97, opened 2016-11-30
utf8 encode auth for jira

macqueen:
fixes https://sentry.io/sentry/sentry/issues/188263407/


src/sentry_plugins/jira/client.py
```diff
@@ -163,7 +163,7 @@ def search_issues(self, project, query):
     def make_request(self, method, url, payload=None):
         if url[:4] != "http":
             url = self.instance_url + url
-        auth = self.username, self.password
+        auth = self.username.encode('utf8'), self.password.encode('utf8')
```
mattrobenolt: To be clear, this is not guaranteed to work since it's possible the data is outside the scope of utf8, so not sure if you care to actually gracefully handle that case or just wait until it happens.

And by gracefully handle, I mean, return some error to the user instead of us 500ing.

PR #104, opened 2016-12-13
[victorops] initial commit

dcramer:


src/sentry_plugins/victorops/plugin.py
```diff
@@ -0,0 +1,93 @@
...
+            for k, v in interface_list
+        ))
+
+    def notify_users(self, group, event, fail_silently=False):
+        if not self.is_configured(group.project):
+            return
+
+        level = event.get_tag('level')
+        if level in ('info', 'debug'):
+            message_type = 'CRITICAL'
```
nategrieb: VictorOps allows a message_type of 'info' so that would make the most sense there.

dcramer:
adding some basic integration tests real quick

dcramer:
this is live on sentry.io

PR #106, opened 2016-12-15
feat(splunk): initial data forwarding plugin

dcramer:


dcramer:
segment will likely land first then we'll refactor things a bit

dcramer:
this is missing some components that im waiting to hear back on

dcramer:
code is functional now

dcramer:
screen shot 2017-12-11 at 12 58 52 pm

src/sentry_plugins/splunk/plugin.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+        host = self.get_host_for_splunk(event)
+        if host:
+            payload['host'] = host
+
+        session = http.build_session()
+        session.post(
+            instance,
+            json=payload,
+            # Splunk cloud instances certifcates dont play nicely
```
mattrobenolt: I feel like this should be a configuration option imo. Seems not great if we're just never verifying SSL since we're sending sensitive data.

dcramer: I dont know if it will ever validate. If we make a configuration option, I have a feeling most people will just shoot themselves in the foot. Alternatively we have to support users uploading their own certificates.

mattrobenolt: Is this the case with Splunk Cloud though? Or just for on-premise hosted Splunk? Does their SaaS spit out invalid certs?

Or is your comment referring to people running their own Splunk?

dcramer: the comment is correct - its splunk cloud thats not verifying with our default certs. On-premise splunk would be an entirely different problem.

mattrobenolt: Oh. Weird. `¯\_(ツ)_/¯`

dcramer:
https://github.com/getsentry/sentry-docs/pull/190

dcramer:
the failure was danger, but only one run (should we just remove danger?)

PR #107, opened 2016-12-15
fix a few jira bugs where custom fields were unecessarily skipped

macqueen:
@getsentry/plugins


dcramer:
lol 🙅‍♂️ 💯

macqueen:
yeah i was thinking this is probably something we should have a test for. i'll add one in another pr.

PR #109, opened 2016-12-20
[segment] update payload to match example spec

dcramer:


dcramer:
fwiw i think library is wrong here, but since integration didn't seem useful it probably fits better

PR #119, opened 2017-01-09
Use yarn to install node_modules

benvinegar:
cc @mattrobenolt @macqueen


.travis.yml
```diff
@@ -7,6 +7,7 @@ services:
 python:
 - '2.7'
 cache:
+  yarn: true
```
benvinegar: This is [supported by Travis](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn). I'll update the other projects to use it.

PR #120, opened 2017-01-09
db: correct hipchat initial migration to create FK fields as bigint

mattrobenolt:
This flat out errors in MySQL since the field in reference is an int
instead of bigint, and the constraint fails to apply.


mattrobenolt:
Got confirmation that this worked on MySQL.

PR #128, opened 2017-02-01
[jira] fix epic name is required showing up erroneously

macqueen:


src/sentry_plugins/jira/static/jira/components/issueActions.jsx
```diff
@@ -16,7 +16,7 @@ class IssueActions extends plugins.DefaultIssuePlugin.DefaultIssueActions {
       state.state = FormState.LOADING;
       this.setState(state, this.onLoad.bind(this, () => {
         this.api.request((this.getPluginCreateEndpoint() +
-                          '?issue_type=' + encodeURIComponent(value)), {
+                          '?issuetype=' + encodeURIComponent(value)), {
```
macqueen: i just changed this so that it would match what jira actually calls this field

src/sentry_plugins/jira/plugin.py
```diff
@@ -152,7 +152,10 @@ def get_new_issue_fields(self, request, group, event, **kwargs):
         # check if the issuetype was passed as a GET parameter
         issue_type = None
         if request is not None:
-            issue_type = request.GET.get('issue_type')
+            if request.method == 'POST':
+                issue_type = request.DATA.get('issuetype')
```
mattrobenolt: Wouldn't this be `request.POST`?

macqueen: no because http://www.tomchristie.com/rest-framework-2-docs/api-guide/requests

PR #131, opened 2017-02-11
[Slack] support optionally including culprit/project fields

stephen:
Hi @dcramer @macqueen, ptal? I would like to be able to not show the culprit/project fields in the Sentry slack notifications.

### DIFF
--- a/src/sentry_plugins/slack/plugin.py
+++ b/src/sentry_plugins/slack/plugin.py
@@ -90,6 +90,20 @@ class SlackPlugin(CorePluginMixin, notify.NotificationPlugin):
             'type': 'bool',
             'required': False,
             'help': 'Include triggering rules with notifications.',
+        }, {
+            'name': 'exclude_project',
+            'label': 'Exclude Project Name',
+            'type': 'bool',
+            'default': False,
+            'required': False,
+            'help': 'Exclude project name with notifications.',
+        }, {
+            'name': 'exclude_culprit',
+            'label': 'Exclude Culprit',
+            'type': 'bool',
+            'default': False,
+            'required': False,
+            'help': 'Exclude culprit with notifications.',
         }]
 
     def color_for_event(self, event):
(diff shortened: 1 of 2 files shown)


stephen:
Hm, travis shows a failure on the build, but the tests look to be passing now.
```
The command "bundle exec danger" exited with 1.
```

src/sentry_plugins/slack/plugin.py
```diff
@@ -90,6 +90,20 @@ def get_config(self, project, **kwargs):
             'type': 'bool',
             'required': False,
             'help': 'Include triggering rules with notifications.',
+        }, {
+            'name': 'include_projects_field',
```
dcramer: for consistency we should make this "include_project"

stephen: Made this change

src/sentry_plugins/slack/plugin.py
```diff
@@ -152,18 +166,18 @@ def notify(self, notification):
 
         # They can be the same if there is no culprit
         # So we set culprit to an empty string instead of duplicating the text
-        if culprit and title != culprit:
+        if self.get_option('include_culprit', project) and culprit and title != culprit:
```
macqueen: both of these will change the behavior for people who have already configured their slack integrations since `self.get_option('include_culprit', project)` will return `None`. can you please adjust your changes so that this won't affect existing slack users?

stephen: Will it be enough to do `self.get_option('include_culprit') == True or self.get_option('include_culprit') == None and ...`? (That seems like it would make the configuration options confusing, with unchecked being either None or False.)

I'm having trouble finding the code that actually calls and handles `get_options`. I had assumed that setting `default` in the configuration would be enough, but I'm not sure when/how that default actually gets applied.

Alternatively - it may be easier to make this opt-in as `exclude_project` and `exclude_culprit` - wdyt?

macqueen: The default only gets applied in the configuration form. The code that calls `get_options` is in the main sentry repo: https://github.com/getsentry/sentry/blob/master/src/sentry/plugins/base/v1.py#L120 which in turn calls https://github.com/getsentry/sentry/blob/master/src/sentry/plugins/helpers.py#L38

I agree it'd probably be simplest to change this to `exclude_project` and `exclude_culprit` and they can default to False.

Thanks!

stephen: Got it - thanks for the pointers! I've updated the PR to s/exclude/include/, and added a couple test cases.

tests/slack/test_plugin.py
```diff
@@ -65,3 +65,79 @@ def test_simple_notification(self):
...
+                    ],
+                    'fallback': '[foo Bar] Hello world',
+                    'title': 'Hello world',
+                    'title_link': 'http://example.com/baz/bar/issues/1/',
+                },
+            ],
+        }
+
+    @responses.activate
+    def test_notification_without_culprit(self):
```
macqueen: i think you want this to be `test_notification_without_project` otherwise 👍

stephen: argh, cargo-culting. fixed!

stephen:
Thanks for the fast response on this! Do you know when this will become available for hosted sentry?

macqueen:
@stephen it's available now

PR #135, opened 2017-02-20
[WIP] Add iTunes Connect plugin

HazAT:


src/sentry_plugins/itunesconnect/__init__.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import absolute_import
+
+from sentry_plugins.base import assert_package_not_installed
+
+assert_package_not_installed('sentry-itunesconnect')
```
mattrobenolt: This isn't needed. This was only for legacy plugins. This package you're referencing never existed.

HazAT:
I will close this PR now since for the foreseeable future we will nit revive it.

PR #144, opened 2017-03-18
[slack] overhaul plugin

dcramer:
Requires https://github.com/getsentry/sentry/pull/5095

 #wip


dcramer:
This doesn't work and is far from done. I'm throwing this up just for a point of reference, as I might scrap this idea.

tl;dr we need oauth to use a lot of newer behavior, and the ideal flow makes the [Slack] button be "Add to Slack" and instantiate the plugin each time it happens. That LIKELY wouldn't use social_auth, as it has annoying constraints.

PR #146, opened 2017-03-23
use repository id instead of repo name in case that changes

macqueen:
not sure if i should also do something like check if repo name is up to date in the `PushEventWebhook`?


src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -36,7 +36,7 @@ def __call__(self, organization, event):
         repo = Repository.objects.get_or_create(
             organization_id=organization.id,
             provider='github',
-            external_id=event['repository']['full_name'],
+            external_id=event['repository']['id'],
```
mattrobenolt: Can we explicitly coerce to a text type here? This is an integer from the API but we're putting this into a charfield.

mattrobenolt:
Let's follow up with keeping `name` in sync since that identifier is used when using the API. So it's probably unituitive if it's always stuck as the wrong name, and there's no UI for them to change it themselves. It'll be stuck in wrong name forever.

dcramer:
The name should be configurable in a future release so there's not much value in writing code to synchronize it unless it's only responding to GitHub renames.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -36,7 +36,7 @@ def __call__(self, organization, event):
         repo = Repository.objects.get_or_create(
```
mattrobenolt: As a part of this, let's change this to just a `get()` and return a 404 when the repository doesn't exist. The lazy creation here is buggy at best anwyays since we can't populate a name. This will allow returnign back something useful to GitHub and a manual re-deliver of failed webhooks can be done.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -33,11 +33,14 @@ class PushEventWebhook(Webhook):
...
-            external_id=event['repository']['full_name'],
-        )[0]
+        try:
+            repo = Repository.objects.get(
+                organization_id=organization.id,
+                provider='github',
+                external_id=six.text_type(event['repository']['id']),
+            )
+        except Repository.DoesNotExist:
+            return 404
```
mattrobenolt: tbh you can just do `raise Http404` like in normal Django and this should just work, without awkwardly returning just a status code.

If you wanted to change the API, I'd suggest we make the handlers return an `HttpResponse` object as a whole.

PR #157, opened 2017-04-03
update compare_commits to take repo model

macqueen:


src/sentry_plugins/github/plugin.py
```diff
@@ -334,8 +334,12 @@ def compare_commits(self, repo, start_sha, end_sha, actor=None):
             raise NotImplementedError('Cannot fetch commits anonymously')
 
         client = self.get_client(actor)
+        # use config name because that is kept in sync via webhooks
+        name = repo.config.get('name')
```
mattrobenolt: We could just `KeyError` here since I don't think we'd wanna fail silently if this were missing?

macqueen: yeah i guess if people go through correct flow, they should definitely have this. changing

PR #162, opened 2017-04-18
try to tie anonymous github users to sentry users

macqueen:
#nochanges relies on https://github.com/getsentry/sentry/pull/5257


src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -51,6 +56,24 @@ def __call__(self, organization, event):
                 author_email = u'{}@localhost'.format(
                     author_email[:65],
                 )
+            # try to figure out who anonymous emails are
+            elif '@users.noreply.github.com' in author_email:
```
mattrobenolt: I'd explicitly do:

```python
elif author_email[-25:] == '@users.noreply.github.com'
```

And work strictly with the assumption that we end with this exact string. This allows avoiding splitting on `@` too. Can then just do:

```python
gh_username = author_email[:-25]
```

This will make it a bit less error prone.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -51,6 +56,24 @@ def __call__(self, organization, event):
...
+                gh_username = author_email.split('@')[0]
+                try:
+                    gh_user = client.request_no_auth('GET', '/users/%s' % gh_username)
+                except ApiError as exc:
+                    logger.exception(six.text_type(exc))
+                else:
+                    try:
+                        user = User.objects.filter(
```
mattrobenolt: Is it possible for us to match more than 1 user? If so, what's our behavior here? Seems bad if we just pick the first one.

macqueen: in theory it is because `UserSocialAuth` is just `unique_together = ('provider', 'uid', 'user')`. that's also why i wanted to filter on organization membership. do you have a better suggestion than picking first? my thought was that as long as they are part of the org, it's probably not a big deal.

PR #164, opened 2017-04-24
[sqs] initial pass on sqs forwarder

dcramer:


src/sentry_plugins/amazon_sqs/plugin.py
```diff
@@ -0,0 +1,70 @@
...
+            region_name=region,
+        )
+
+        client.send_message(
+            # TODO(dcramer): Amazon doesnt support payloads larger than 256kb
+            # We could support this by simply trimming it and allowing upload
+            # to S3
+            QueueUrl=queue_url,
+            MessageBody=json.dumps(payload),
+            MessageDeduplicationId=event.event_id,
```
mattrobenolt: We need to drop this since this is specifically for FIFO queues which seem pretty unsupported all around.

src/sentry_plugins/amazon_sqs/plugin.py
```diff
@@ -0,0 +1,70 @@
...
+
+        client = boto3.client(
+            service_name='sqs',
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+            region_name=region,
+        )
+
+        client.send_message(
+            # TODO(dcramer): Amazon doesnt support payloads larger than 256kb
```
mattrobenolt: We should also explicitly capture this exception. boto yells pretty loudly when this hits because Amazon is returning back HTML with a 413 response.

This is wrapped in `safe_execute`, but imo this could just be a lot of noise in Sentry with something that we can't even fix.

Alternatively, we just do like,

```python
message = json.dumps(payload)
if len(message) > 256*1024:
  return
```

And just don't even attempt to send.

setup.py
```diff
@@ -31,6 +31,8 @@
 
 install_requires = [
     'BeautifulSoup>=3.2.1',
+    # sentry also requires this, so we're just enforcing that it needs to exist
+    'boto3>=1.4.4,<1.5.0',
```
mattrobenolt: Just heads up, I don't see why you'll require 1.4.4. We have it pinned to 1.4.1 in getsentry today, so we can either loosen this up here, or bump it in getsentry or we'll be conflicting.

dcramer: something was broken on 1.4.1 IIRC (it might have been the messageduplicationid)

PR #165, opened 2017-04-25
actually save external id to commit author

macqueen:
realized i wasn't actually saving this to prevent future requests


src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -99,7 +99,10 @@ def __call__(self, organization, event):
                                 author_email = user.email
                                 gh_username_cache[gh_username] = author_email
                                 if commit_author is not None:
-                                    commit_author.update(email=author_email)
+                                    commit_author.update(
+                                        email=author_email,
+                                        external_id=gh_username,
```
mattrobenolt: Hmm, I wonder if we should specify the provider in the `external_id`. Something like: `'github:%s' % gh_username`. I feel like there'd be a problem when we have multiple providers with the same username across them.

macqueen: @mattrobenolt yeah i was kinda thinking about that, but `CommitAuthor` has a unique constraint on organization id and external id, so seems pretty unlikely?

mattrobenolt: Well, to be fair, imagine if used say, both GitHub and Phabricator (hypothetically). And I have the same username on both systems.

macqueen: why is that a big deal? we'd think you were the same author, which you are. i think the case where it would be bad is where you're something in github and then i have the same username that you do in github in phabricator.

mattrobenolt: Let's flip it. Say I got the username in our private Phabricator as `matt`, then on GitHub `matt` is definitely not me and refers to someone else. I just don't see the harm in adding a prefix for the provider to just avoid the collision in the future. It'd save a migration in the future if we hit this problem.

macqueen: what if your phabricator username was `github:matt` :trollface: 

jk will update it.

mattrobenolt: > what if your phabricator username was github:matt :trollface:

This wouldn't matter since we'd be storing it as `phabricator:github:matt` instead. :)

PR #167, opened 2017-04-26
don't call response.json when no content fixes SENTRY-36S

macqueen:


src/sentry_plugins/github/client.py
```diff
@@ -28,7 +28,9 @@ def _request(self, method, path, headers=None, data=None, params=None):
             resp.raise_for_status()
         except HTTPError as e:
             raise ApiError.from_response(e.response)
-        return resp.json()
+
+        if resp.status_code != 204:
```
mattrobenolt: Let's change this to:

```python
if resp.status_code == 204:
    return {}
return resp.json()
```

Just to avoid returning `None`

PR #170, opened 2017-05-01
Support either of Sentry's URL patterns for HipChat card expansion.

RussellRollins:
We have recently been setting up Sentry's HipChat Integration. One issue we found was that link expansion was not working correctly.

After digging around, we determined that that HipChat plugin expects URLs with a group parameter, but our Sentry is producing URLs with an issues parameter instead.

We were able to produce successful card expansions by simply swapping the "issues" in our URLs with "group".

It may make sense to rename this parameter throughout, but the simplest, most immediately backward compatible fix seems to be to simply allow either parameter in the link regex.

### DIFF
--- a/src/sentry_plugins/hipchat_ac/views.py
+++ b/src/sentry_plugins/hipchat_ac/views.py
@@ -52,7 +52,7 @@ def get_link_regexp():
     pattern = get_link_pattern()
     if pattern not in _regexp_cache:
         _regexp_cache[pattern] = re.compile(pattern +
-            r'(?P<org>[^/]+)/(?P<proj>[^/]+)/group/'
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Oh man, we renamed this a while back. Thanks for the fix!

mattrobenolt:
Looks like you have a failing test. :)

RussellRollins:
Surprisingly, I wrote a good regex but a bad test, that's the opposite of what I would have expected.

This is passing tests now, I'm uncertain what:

```Failing the build, Danger cannot run without API access.```

Means though.

mattrobenolt:
You're all good. :)

RussellRollins:
Excellent. Thanks for the quick response. Love what y'all are doing with Sentry.

mattrobenolt:
🤗

PR #172, opened 2017-05-03
Get commits without start sha

kjlundsgaard:
Allows `compare_commits` to be called when `start_sha` is None and calls the Github API to fetch most recent commits based on `end_sha`


src/sentry_plugins/github/plugin.py
```diff
@@ -336,8 +336,17 @@ def compare_commits(self, repo, start_sha, end_sha, actor=None):
...
-        except Exception as e:
-            self.raise_error(e)
-        return [{'id': c['sha'], 'repository': repo.name} for c in res['commits']]
+        if start_sha is None:
+            try:
+                res = client.get_last_commits(name, end_sha)
+            except Exception as e:
+                self.raise_error(e)
+            else:
+                return [{'id': c['sha'], 'repository': repo.name} for c in res]
```
macqueen: let's maybe limit this to 10 commits for now? kind of arbitrary, but i'm thinking of this mostly as a way for people to make sure their setup is correct

PR #186, opened 2017-05-19
WIP sort of figure out flow of github integrations installation

macqueen:


src/sentry_plugins/github/views.py
```diff
@@ -0,0 +1,74 @@
...
+            form = OrganizationForm(
+                [(o.id, o.name) for o in request.user.get_orgs()],
+                request.POST,
+            )
+            if form.is_valid():
+                # TODO(jess): this isn't actually working, but
+                # we need to somehow verify user has access
+                # to this installation
+
+                # auth = UserSocialAuth.objects.filter(
```
macqueen: i need to figure out why this wasn't working, but we need to do something similar to make sure the user actually has permission to access this integration

macqueen:
gonna close this so you don't get spammed as i update this branch

PR #188, opened 2017-05-26
support github integrations

macqueen:


dcramer:
With a big PR like this it'd be helpful to have a general description of how things should work

macqueen:
@dcramer i put some specific questions in the description of https://github.com/getsentry/sentry/pull/5461. I need to answer those / implement whatever we decide on before I need a more thorough review.

conftest.py
```diff
@@ -71,3 +71,4 @@ def pytest_configure(config):
     settings.BITBUCKET_CONSUMER_SECRET = '123'
     settings.GITHUB_APP_ID = 'abc'
     settings.GITHUB_API_SECRET = '123'
+    settings.GITHUB_INTEGRATION_HOOK_SECRET = 'b3002c3e321d4b7880360d397db2ccfd'
```
mattrobenolt: 😐 

Going to assume this isn't a valid token. If it is, can we revoke? If not, can we comment that it's not real?

src/sentry_plugins/github/client.py
```diff
@@ -112,26 +147,79 @@ def delete_hook(self, repo, id):
...
+            self.token = res['token']
+            self.expires_at = datetime.datetime.strptime(
+                res['expires_at'],
+                '%Y-%m-%dT%H:%M:%SZ',
+            )
+
+        return self.token
+
+    def get_jwt(self):
+        private_key = settings.GITHUB_INSTALLATION_PRIVATE_KEY
```
mattrobenolt: This doesn't appear to be set anyways, even as a default value. So I assume things will blow up since this will be an `AttributeError`.

I'd suggest making this a check at the module level.

Something like:

```python
GITHUB_INSTALLATION_PRIVATE_KEY = getattr(settings, 'GITHUB_INSTALLATION_PRIVATE_KEY', None)
```

Maybe even just raising a RuntimeError here if we don't have anything set.

src/sentry_plugins/github/client.py
```diff
@@ -112,26 +147,79 @@ def delete_hook(self, repo, id):
...
+        exp = datetime.datetime.utcnow() + datetime.timedelta(minutes=10)
+        exp = calendar.timegm(exp.timetuple())
+        # Generate the JWT
+        payload = {
+            # issued at time
+            'iat': int(time.time()),
+            # JWT expiration time (10 minute maximum)
+            'exp': exp,
+            # Integration's GitHub identifier
+            'iss': settings.GITHUB_INTEGRATION_APP_ID,
```
mattrobenolt: Same applies here.

We could/should alternatively move these to `options` instead of settings.

conftest.py
macqueen: not real! it was one we were already using in the tests: https://github.com/getsentry/sentry-plugins/blob/master/tests/github/endpoints/test_webhooks.py#L88

but i'll add a comment to make that clearer

src/sentry_plugins/github/plugin.py
```diff
@@ -399,3 +409,141 @@ def compare_commits(self, repo, start_sha, end_sha, actor=None):
...
+
+        # check that user actually has access to add
+        allowed_gh_installations = set(self.get_installations(user))
+        if int(integration.external_id) not in allowed_gh_installations:
+            raise PluginError('You do not have access to that integration')
+
+        integration.add_organization(organization.id)
+
+        for repo in self.get_repositories(integration):
+            # TODO(jess): figure out way to migrate from github --> github apps
```
macqueen: it wouldn't be difficult to do this, just wondering if we should, and if so, if i should also delete the individual repo's webhook in favor of using the integration's

macqueen:
@dcramer if you wanna look at this again, i think it's ready for us to start using it.

src/sentry_plugins/github/__init__.py
```diff
@@ -1,5 +1,29 @@
...
+    'plugins.github.integration-hook-secret',
+    default='',
+    flags=FLAG_REQUIRED | FLAG_PRIORITIZE_DISK,
+)
+register(
+    'plugins.github.integration-app-id',
+    default=0,
+    flags=FLAG_REQUIRED | FLAG_PRIORITIZE_DISK,
+)
+register(
```
dcramer: i think all of these aren't required, right?

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -35,13 +37,67 @@ def get_external_id(username):
...
+                            'url': 'https://github.com/%s' % (r['full_name'], ),
+                            'config': config,
+                            'integration_id': integration.id,
+                        }
+                    )
+                    if not created:
+                        repo.config.update(config)
+                        repo.integration_id = integration.id
+                        repo.save()
+        # TODO(jess): what do we want to do when they're removed?
```
dcramer: def seems like something we can be reactive too

e.g. we tried querying this data, and mark an integration as 'suspended' or something in the UI (we can follow up with this later)

dcramer:
My main concern at this point is the sheer volume of code that isn't run through any kind of automation. I also wonder if we need to abstract all the followup TODOs out into a ticket so we don't forget about them.

getsentry-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">You should update CHANGES due to the size of this PR</td>
    </tr>
  </tbody>
</table>

</p>

PR #200, opened 2017-06-29
Slack Plugin - Remove unsupported parameter 'parse'

timmygee:
Slack doesn't seem to support the `parse` parameter in the webhook payload. Removing it enables the plugin to work with slack again

### DIFF
--- a/src/sentry_plugins/slack/plugin.py
+++ b/src/sentry_plugins/slack/plugin.py
@@ -214,7 +214,6 @@ class SlackPlugin(CorePluginMixin, notify.NotificationPlugin):
                 })
 
         payload = {
-            'parse': 'none',
             'attachments': [{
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Hmm, weird, does this mean it 100% doesn't work right now? When did this change in their API?

mattrobenolt:
I don't see any time of when this was removed, but it's definitely not documented in their API today. And I added this field 3 years ago. https://github.com/getsentry/sentry-slack/commit/bde30a1f32ecc391c93fa683ef76251af3225d55

So this seems fine to me.

mattrobenolt:
Thanks! 🍕

timmygee:
Welcome! Yeah I could not find any evidence of it being removed either, it just didn't work anymore!

PR #213, opened 2017-08-03
make pagerduty url overridable with environment variable

yuekui2:
for testing purpose, we need to a way to override the official pagerduty api integration url with a customized one specified by an environment variable.


yuekui2:
Hi Folks,

Any concern with this CL? If not, please help assign someone review and get it in. It has been a while for this pull request. 

Thanks.
-Kui

Issue #215, opened 2017-08-07
[github] assignee list only includes first 30 users


macqueen:
see https://developer.github.com/v3/#pagination

PR #221, opened 2017-08-21
increase number of github assignees to 100

macqueen:
this is a temporary fix for https://github.com/getsentry/sentry-plugins/issues/215 until someone has time to come up with a better idea


rkrzr:
Thanks for fixing this! This feature is very convenient.

> until someone has time to come up with a better idea

Perhaps you could just iterate over all available pages and return all members of the organization?

PR #230, opened 2017-09-12
fix(github): Add logging and improve error message when repo creation fails

macqueen:
partially addresses https://github.com/getsentry/sentry/issues/6096


src/sentry_plugins/github/plugin.py
```diff
@@ -343,6 +350,11 @@ def create_repository(self, organization, data, actor=None):
                 }
             )
         except Exception as e:
+            self.logger.exception('Unable to create webhook %s', six.text_type(e))
```
dcramer: could we do something like ``github.webhook.create-failure`` for the msg (to match others)

(also i dont care so much about the event name, but something reasonable)

also what matt said about doing ``extra={}``, adding in the status code, and adding in the organization and project slug, as well as possibly the repo name?

PR #233, opened 2017-09-14
remove breaking code

brianwu02:
Hey. Unless I'm mis-understanding how this feature works, it looks like this single line of code fails because it's not referenced anywhere and breaks the entire plugin. Tested, works fine after removing this line of code.


mattrobenolt:
This requires version 8.20 of the server. sentry-plugins version just be paired with the correct server version. See https://github.com/getsentry/sentry/pull/5881

mattrobenolt:
To clarify, my guess is you’re just running a newer version of plugins. Maybe like, 8.20 of plugins, but only 8.19 of the server. They need to always be the same version and are released in tandem.

brianwu02:
ahhh. Thank you!

PR #243, opened 2017-09-18
fix(jira): Pass create meta params correctly

macqueen:


dcramer:
this can be closed as it was superceded by a diff fix

PR #245, opened 2017-09-20
[sessionstack] Fixed access token generation bug

lsnickolov:


dcramer:
This broke the tests. I'm confused as how this would be correct. Wouldn't you want to get the access token first, and then create it if it doesnt exist?

PR #246, opened 2017-09-20
fix(sessionstack): Correct failing test

dcramer:


dcramer:
This is for the change in behavior introduced in #245

dcramer:
This isnt quite a fix

dcramer:
this 'fixes' it, though i think the code is still wrong

PR #250, opened 2017-09-25
ref(vsts): Handle 203 error in a more consistent manner

dcramer:
Refs getsentry/sentry#6184


src/sentry_plugins/vsts/client.py
```diff
@@ -1,29 +1,36 @@
...
+            # XXX(dcramer): I'm not entirely convinced this is an authentication
+            # error, but we've seen in the past that refreshing the token would
+            # resolve this response, and unfortunately there's not much detail
+            # online about it.
+            # https://github.com/getsentry/sentry/issues/5263
+            if exc.code == 203:
+                six.reraise(
+                    ApiUnauthorized,
+                    ApiUnauthorized('Received an HTTP %s %s response from the API.', exc.code),
+                    sys.exc_info[2],
```
MaxBittker: TIL. does this need to be `sys.exc_info()[2]`?

dcramer: good catch

PR #251, opened 2017-09-25
fix(vsts): Correct and improve project selection

dcramer:
Fix incorrectly only using the default_project config. Additionally add basic prepopulated choices.


dcramer:
http://i.cra.mr/0Y1Q2d213242

http://i.cra.mr/2H3V0z181m31

dcramer:
Aside, this also changes the content type back to the patch format, which seems to be required. I thought I had tested it with just application/json, but I must not have.

src/sentry_plugins/vsts/client.py
```diff
@@ -20,7 +20,7 @@ def exception_means_unauthorized(self, exc):
     def request(self, method, path, data=None, params=None):
         headers = {
             'Accept': 'application/json; api-version={}'.format(self.api_version),
-            'Content-Type': 'application/json',
+            'Content-Type': 'application/json-patch+json',
```
dcramer: also, given this (I think) means 'application/json' and 'application/json-patch', our content type detection should probably use a library that does this for us, or should be improved

PR #255, opened 2017-10-02
feat(vsts): Correct naming for VSTS

dcramer:


dcramer:
Shorthand name requires https://github.com/getsentry/sentry/pull/6239

PR #260, opened 2017-10-03
fix(vsts): Correct error propgation on issue creation

dcramer:


dcramer:
I think we'll likely pull in the core error handling as this is getting a little silly

dcramer:
Also opened up https://github.com/getsentry/sentry/pull/6250 to avoid excess code in this repo

dcramer:
shipping this as its minimal impact but resolves some user facing concerns

PR #261, opened 2017-10-03
fix(bitbucket): Handle OAuth1 correctly on invalid tokens

dcramer:


dcramer:
I didn't actually test this, and unfortunately theres not a lot of automated testing in place.

MaxBittker:
this seems to have failed the tests in case you didn't see

dcramer:
Tests should be good now.

PR #264, opened 2017-10-10
ref: Use tagstore abstraction in Slack plugin

bretthoerner:
This requires Sentry `master`, I assume it'll fail if we manually pin this repo (seems like it?).


mattrobenolt:
> I assume it'll fail if we manually pin this repo (seems like it?).

Yeah, we can just merge the sentry PR, then re-run theses here to double check.

PR #269, opened 2017-10-18
fix: Disallow redirects by default on non-GET API calls

dcramer:
With vsts we were hitting a 302, which was causing excess confusion on what the (still invalid) response was.


dcramer:
@macqueen @MaxBittker are you aware of any integrations that were relying on redirects?

dcramer:
Refs getsentry/sentry#6184

macqueen:
@dcramer yeah i think github did

macqueen:
https://github.com/getsentry/sentry-plugins/pull/67

dcramer:
Ok I'll enable for GitHub, and we can reactively enable case-by-case after that.

dcramer:
Improved behavior to only enable redirects on GET by default, and to enable fully on GitHub

mattrobenolt:
> Improved behavior to only enable redirects on GET by default, and to enable fully on GitHub

To be fair, most things shouldn't follow redirects on say, POST. Actually surprised if this is the behavior internally. :/

dcramer:
@mattrobenolt it was PATCH, so could always be a flaw in requests

PR #273, opened 2017-10-24
fix(jira): Make sure base instance url is used for user picker

macqueen:
this should fix the case where the instance url is something like `https://sentry.io/jira/`

cc @MeredithAnya


billyvg:
What's an example `jira_url`?

macqueen:
@billyvg `https://domain.atlassian.net/rest/api/1.0/users/picker?fieldName=customfield_1111111&showAvatar=true&query=`

PR #274, opened 2017-10-25
fix(bitbucket): Update incoming IP ranges

ChrisOgden:
See new IPs here: https://blog.bitbucket.org/2017/06/21/new-outbound-ip-addresses-webhooks/

### DIFF
--- a/src/sentry_plugins/bitbucket/endpoints/webhook.py
+++ b/src/sentry_plugins/bitbucket/endpoints/webhook.py
@@ -22,7 +22,8 @@ logger = logging.getLogger('sentry.webhooks')
 
 # Bitbucket Cloud IP range:
 # https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html#Managewebhooks-trigger_webhookTriggeringwebhooks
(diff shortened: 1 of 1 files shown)


src/sentry_plugins/bitbucket/endpoints/webhook.py
```diff
@@ -22,7 +22,7 @@
 
 # Bitbucket Cloud IP range:
 # https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html#Managewebhooks-trigger_webhookTriggeringwebhooks
-BITBUCKET_IP_RANGE = ipaddress.ip_network(u'104.192.143.0/24')
+BITBUCKET_IP_RANGE = [ipaddress.ip_network(u'104.192.143.0/24'), ipaddress.ip_address('34.198.203.127'), ipaddress.ip_address('34.198.178.64'), ipaddress.ip_address('34.198.32.85')]
```
gobengo: the new ip_address() values need to be like `u'34...127'` (unicode strings)

MaxBittker:
thanks for the pr @ChrisOgden !  would you mind making @gobengo 's formatting suggestion, and (not 100% necessary, but if you're touching the file anyway) 
factoring `ipaddress.ip_address(six.text_type(request.META['REMOTE_ADDR'])` into it's own variable so it's not repeated? This totally flew under my radar so I really appreciate the vigilance,

MaxBittker:
thanks for the contribution chris, we actually fixed this in a separate PR

ChrisOgden:
Great, sorry I couldn't get back to this.

PR #281, opened 2017-11-17
feat(Github): Track Pull Requests

MaxBittker:
Collect pull request information from github webhooks, handles two actions: opened & edited. 
It does not keep track of the commits ("sync" events).

possible concern: upgrading the existing webhook scopes, this will need to be solved before merging! solution: write a migration to update them and roll forward

NOTE: unbump sentry before merging 

needs to merge after: https://github.com/getsentry/sentry/pull/6635


getsentry-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is classed as Work in Progress</td>
    </tr>
  </tbody>
</table>

</p>

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,120 @@ def __call__(self, event, organization=None):
...
+            self._handle_created(event, organization, repo, is_apps)
+        if action == 'edited':
+            self._handle_updated(event, organization, repo)
+        if action == 'synchronize':
+            self._handle_sync(event, organization, repo)
+
+    def _handle_created(self, event, organization, repo, is_apps):
+        """PR was created"""
+
+        client = GitHubClient()
```
macqueen: i think you're going to need to pass auth into this for any private repos (and we'd want to for public ones due to api rate limits). since this is a webhook and there isn't a user making the request (so it's unclear which auth we should use) we're probably going to have to think of a work around.

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,120 @@ def __call__(self, event, organization=None):
...
+
+    def _handle_updated(self, event, organization, repo):
+        """PR title or description was edited"""
+
+        pull_request = event['pull_request']
+        number = pull_request['number']
+        title = pull_request['title']
+        body = pull_request['body']
+
+        pr = ChangeRequest.objects.get(
```
macqueen: you should handle a `ChangeRequest.DoesNotExist` error

macqueen: you also could probably combine this and the update to a single query by doing:
```python
ChangeRequest.objects.filter(
    repository_id=repo.id,
    key=number,
).update(
    title=title,
    message=body,
)

```

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,120 @@ def __call__(self, event, organization=None):
...
+
+        pr.update(
+            title=title,
+            message=body,
+        )
+
+    def _handle_sync(self, event, organization, repo):
+        """New Commits have been pushed"""
+
+        client = GitHubClient()
```
macqueen: same issue with auth here

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,120 @@ def __call__(self, event, organization=None):
...
+
+        pull_request = event['pull_request']
+        number = pull_request['number']
+
+        pr = ChangeRequest.objects.get(
```
macqueen: same comment re: `DoesNotExist`/combining into a filter/update query

src/sentry_plugins/github/plugin.py
```diff
@@ -316,7 +316,7 @@ def create_repository(self, organization, data, actor=None):
                 data['name'], {
                     'name': 'web',
                     'active': True,
-                    'events': ['push'],
+                    'events': ['push', 'pull_request'],
```
macqueen: Not totally sure what the best way to handle existing repos is. There's an api for updating webhooks: https://developer.github.com/v3/repos/hooks/#edit-a-hook, but we need to know which user auth to use, and apart from having something in the Sentry UI that prompts users to "upgrade" to include PRs, I don't know how we'd figure out which user to use. The other thing we could do is just tell users with existing repos to update the events in the Github UI, but probably some people will do that incorrectly. This seems like something Matt might have an opinion on, so let's ask him.

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: we do the same thing in commit i think, why does it work there? could we store auth on the repo?

macqueen: that's just the user endpoint, which doesn't require auth: https://github.com/getsentry/sentry-plugins/blob/master/src/sentry_plugins/github/endpoints/webhook.py#L157

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: did you mean get or filter here? those are unique together keys i think

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: how do i combine this one? it's a many2many

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: dang

MaxBittker:
@macqueen I think this is ready for re-review, but i'm still not sure what to do when update the permissions.

src/sentry_plugins/github/plugin.py
MaxBittker: this is still an unresolved problem in the PR

MaxBittker: i wrote a script for this btw

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,87 @@ def __call__(self, event, organization=None):
...
+
+        # We need to track GitHub's "full_name" which is the repository slug.
+        # This is needed to access the API since `external_id` isn't sufficient.
+        if repo.config.get('name') != event['repository']['full_name']:
+            repo.config['name'] = event['repository']['full_name']
+            repo.save()
+
+        if action == 'opened':
+            self._handle_created(event, organization, repo, is_apps)
+        if action == 'edited' or action == "closed":
```
macqueen: nit, should be single quotes on `"closed"`

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,87 @@ def __call__(self, event, organization=None):
...
+            self._handle_updated(event, organization, repo)
+
+    def _handle_created(self, event, organization, repo, is_apps):
```
macqueen: it doesn't look like `is_apps` is used

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,87 @@ def __call__(self, event, organization=None):
...
+        """PR was created"""
+
+        pull_request = event['pull_request']
+        number = pull_request['number']
+        title = pull_request['title']
+        body = pull_request['body']
+        user = pull_request['user']
+        merge_commit_sha = pull_request['merge_commit_sha']
+
+        # get author (TODO: use more complicated logic from the commit webhook)
```
macqueen: does this need to be done before this is merged?

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,87 @@ def __call__(self, event, organization=None):
...
+            PullRequest.objects.get(
+                repository_id=repo.id,
+                key=number,
+            ).update(
+                title=title,
+                message=body,
+                merge_commit_sha=merge_commit_sha,
+            )
+        except PullRequest.DoesNotExist:
+            pass
```
macqueen: if you're not going to do anything as a result of the `PullRequest.DoesNotExist` i'm pretty sure you could make this a single query instead of two by using `filter` instead of `get`

tests/github/endpoints/test_webhooks.py
```diff
@@ -5,13 +5,20 @@
 
 from datetime import datetime
 from django.utils import timezone
-from sentry.models import (Commit, CommitAuthor, Integration, OrganizationOption, Repository)
+from sentry.models import (
+    PullRequest,
```
macqueen: nit, i think these are supposed to be in alphabetical order

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -273,9 +273,87 @@ def __call__(self, event, organization=None):
...
+        # get author (TODO: use more complicated logic from the commit webhook)
+        author = author = CommitAuthor.objects.get_or_create(
+            organization_id=organization.id,
+            external_id=user['id'],
+            defaults={
+                'name': user['login']
+            }
+        )[0]
+
+        PullRequest.objects.create(
```
macqueen: you should probably handle an `IntegrityError` here

PR #290, opened 2017-12-12
fix(vsts): Add vsts to tests

dcramer:


dcramer:
this was always wrong, but was only breaking the selenium tests

PR #302, opened 2018-01-17
resolve a reasonable author email

MaxBittker:
Throw the kitchen sink at finding an author email

fixes SENTRY-5QK


src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -310,10 +310,28 @@ def _handle_created(self, event, organization, repo, is_apps):
...
-        # get author TODO(Maxbittker): do we need to use the more complicated
-        # logic from the commit webhook?
-        author = author = CommitAuthor.objects.get_or_create(
+        author_email = u'{}@localhost'.format(user['login'])
+        try:
+            commit_author = CommitAuthor.objects.get(
+                external_id=get_external_id(user['login']),
+                organization_id=organization.id,
+            )
+            author_email = commit_author.email
```
macqueen: nit, but i'd put this line in an `else` block since it's not going to cause a `DoesNotExist`

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -310,10 +310,28 @@ def _handle_created(self, event, organization, repo, is_apps):
         user = pull_request['user']
         merge_commit_sha = pull_request['merge_commit_sha']
 
-        # get author TODO(Maxbittker): do we need to use the more complicated
-        # logic from the commit webhook?
-        author = author = CommitAuthor.objects.get_or_create(
+        author_email = u'{}@localhost'.format(user['login'])
```
macqueen: mmm why localhost and not trying to match the github noreply emails?

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -310,10 +310,28 @@ def _handle_created(self, event, organization, repo, is_apps):
...
+        try:
+            commit_author = CommitAuthor.objects.get(
+                external_id=get_external_id(user['login']),
+                organization_id=organization.id,
+            )
+            author_email = commit_author.email
+        except CommitAuthor.DoesNotExist:
+            try:
+                user = User.objects.filter(
```
macqueen: i think your redefinition of user here actually might cause a bug cause below you do `external_id=user['id']` and `user['login'][:128]`

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -310,10 +310,28 @@ def _handle_created(self, event, organization, repo, is_apps):
...
+                    social_auth__provider='github',
+                    social_auth__uid=user['id'],
+                    org_memberships=organization,
+                )[0]
+            except (IndexError, KeyError):
```
macqueen: will there ever actually be a keyerror here?

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: emulating https://github.com/getsentry/sentry-plugins/blob/fix/pr/commit_author/src/sentry_plugins/github/endpoints/webhook.py#L130-L132
(and what's in the db already)

src/sentry_plugins/github/endpoints/webhook.py
MaxBittker: goood catch

PR #309, opened 2018-01-26
 fix(Bitbucket): guard key error in commit fetching

MaxBittker:
fixes SENTRY-5SZ
fixes SENTRY-5T0

don't try to page when there's no next link


MaxBittker:
thanks matt.

PR #311, opened 2018-01-29
fix(sessionstack): Bind click handler

MaxBittker:
fixes JAVASCRIPT-2W5


src/sentry_plugins/sessionstack/static/sessionstack/contexts/sessionstack.jsx
```diff
@@ -49,31 +49,37 @@ class SessionStackContextType extends React.Component {
...
+            src={session_url}
+            sandbox="allow-scripts allow-same-origin"
+            width={this.state.width}
+            height={this.state.height}
+          />
+        ) : (
+          <button
+            className="btn btn-default"
+            type="button"
+            onClick={() => this.playSession()}
```
MaxBittker: this is the only real diff, sorry for the prettiering

billyvg:
Can we move our `.prettierrc` into this repo?

MaxBittker:
good idea.

PR #319, opened 2018-02-12
WIP: Add GitLab support for Sentry Releases

stanhu:


uda:
@jamincollins since it doesn't create a webhook it doesn't have to delete it when removed.

jamincollins:
> @jamincollins since it doesn't create a webhook it doesn't have to delete it when removed.

In testing, the repository was never removed after adding a GitLab repository and requesting that the repository be removed.  So, I thought the issue was that it doesn't have deletion logic.

uda:
@jamincollins I think that is a different issue, I think so.

I looked at other implementations, and it seems they simply act on the external service, if that is required, not on the sentry DB.

Github plugin:
https://github.com/getsentry/sentry-plugins/blob/3b08a43ea9ca1fb0fd183c3fe7bd5606f14ba993/src/sentry_plugins/github/plugin.py#L388-L399

Bitbucket plugin:
https://github.com/getsentry/sentry-plugins/blob/3b08a43ea9ca1fb0fd183c3fe7bd5606f14ba993/src/sentry_plugins/bitbucket/repository_provider.py#L97-L107

jamincollins:
> @jamincollins I think that is a different issue, I think so.
> 
> I looked at other implementations, and it seems they simply act on the external service, if that is required, not on the sentry DB.

I'm talking about in the Sentry UI.  With the existing plugin changes, one **can** add a GitLab repository.  However, they **can not** remove it once created.  Attempting to remove it leaves it perpetually pending removal, based on my experience and testing.  As such, once created it **can not** be changed or recreated (should you need to change the access token for example, which was my case).

I've also yet to find any benefit of adding a GitLab repository using this plugin.  Granted I don't know what the benefit _should_ be as I've never tried any of the other repository plugins with Sentry either.  I thought linking a repository would allow the stack traces in the reports to link to the relevant code, but if that is supposed to happen, it has not been my experience thus far.

MeredithAnya:
Just a heads up, support for Gitlab (self-managed version) repositories is in the process of being moved into core Sentry, along with issue tracking. This plugin will still exist - for the time being - but will be limited to it's current functionality (creating/linking issues), we won't be adding to it. 

Appreciate the work so far though! (A lot of it will look pretty similar to what you have)

bsameh:
@MeredithAnya What about the cloud version of GitLab? Do you mind sharing whether there are any plans to directly integrate it with Sentry (similar to the recent integrations with GitHub and VSTS)?

zerthimon:
@MeredithAnya This is only for the SaaS sentry, will this be available for the on-premise versions ?

zerthimon:
@markstory Awesome, thanks!

PR #321, opened 2018-02-13
Ops 1745 Patch sentry-plugins

alexef:
As sentry-jira is deprecated, let's port our changes to the currently supported plugins.


alexef:
My bad, PR was not meant for upstream sentry.

Issue #105, opened 2016-12-15
Add clubhouse.io integration


benvinegar:
https://clubhouse.io

cannikin:
I'm surprised you guys haven't got more requests for this one, Clubhouse is the new hotness in project management! Starting to see stores of people leaving Jira to come over here.

Is plugin development open source in any way? Can I patch something together and open a pull request?

benvinegar:
So, obviously this whole repository is open source, and there are instructions for checking out the main Sentry project, checking out sentry-plugins, and developing a new plugin. Many of our plugins today were developed this way.

The problem is that if we officially endorse such a plugin (and make it available via sentry.io), we're now responsible for it, and it's not typical that the original contributors stick around to make sure bug fixes and updates are made for other users. If the plugin breaks (e.g. because Clubhouse changes their API) – it's now on us, and if we don't deliver users become upset.

devteamexpress:
+1

BenCoffeed:
+1

cannikin:
I'd happy write it myself but I'm not a Python guy. :( Anyone else?

BenCoffeed:
I've opened a support request with ClubHouse and they mentioned that they use Sentry internally and there is interest in a Sentry plugin there as well. So, 🤞 we'll see something from the Clubhouse side.

aledoo:
+1000

flxfxp:
+10000

ste00martin:
I want this

mmartinson:
How does sentry.io make business decisions about which plugins to officially support? I would make the case that clubhouse could be a high value integration to bring/keep paying customers.

I asked Clubhouse about it and their CS lead told me "I've been in touch with the team at Sentry and their product and engineering team would appreciate comments on this Issue <linked to this issue>"

esauro:
This would be a fantastic thing to have. Absolutely desirable. Please bring clubhouse support to sentry.io ;)

bendavies:
Echoing the above. Would love to see clubhouse integration.

jamtur01:
Ditto - would like to see this.

miketheman:
Would the sentry.io team accept a PR for this and work it's way into the release cycle?

mattrobenolt:
[...]

PR #324, opened 2018-02-24
Add Clubhouse plugin as Issue Tracker

miketheman:
Refs #105

Initial implementation of a Clubhouse Story plugin.

See inline for notes of open questions.

I welcome your feedback, as I've not written a Sentry plugin before, and I tried to do the things that made most sense while reading other plugins.

# Screenshots

Expand to see the goodies.

<details>

## Settings page

Invalid Project id:

screen shot 2018-02-24 at 11 27 24 am

Valid project id:

screen shot 2018-02-24 at 11 16 15 am

## Event page

Create new Story:

screen shot 2018-02-24 at 11 21 41 am

New Story input form:

screen shot 2018-02-24 at 11 21 53 am

Resulting Event link:

screen shot 2018-02-24 at 11 16 30 am

## Clubhouse Page

screen shot 2018-02-24 at 11 22 29 am

</details>


getsentry-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Changes to build requirements</td>
    </tr>
  </tbody>
</table>

</p>

src/sentry_plugins/clubhouse/plugin.py
```diff
@@ -0,0 +1,95 @@
...
+from sentry.plugins.bases.issue2 import IssuePlugin2, IssueGroupActionEndpoint
+
+from sentry_plugins.base import CorePluginMixin
+from sentry_plugins.utils import get_secret_field_config
+
+from .client import ClubhouseClient
+
+
+class ClubhousePlugin(CorePluginMixin, IssuePlugin2):
+    # TODO: There currently isn't a good "list all story titles" endpoint for Clubhouse.
```
jamtur01: @miketheman I've logged this with Clubhouse.

src/sentry_plugins/clubhouse/plugin.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return response['id']
+
+    def get_issue_label(self, group, issue_id, **kwargs):
+        return 'Clubhouse Story #%s' % issue_id
+
+    def get_issue_url(self, group, issue_id, **kwargs):
+        # TODO: Uncertain of this is a good method, as it incurs an API call on page load.
+        # `issue_url` can't be generated purely from a knonwn string + `issue_id`, as it
+        # requires the Clubhouse Organization in the URL, which isn't currently a known API call.
```
jamtur01: And this too.

src/sentry_plugins/clubhouse/plugin.py
```diff
@@ -0,0 +1,95 @@
...
+        client = ClubhouseClient()
+        story = client.get_story(issue_id, token=self.get_option('token', group.project))
+        return story['app_url']
+
+    def validate_config(self, project, config, actor):
+        try:
+            config['project'] = int(config['project'])
+        except ValueError as exc:
+            self.logger.exception(six.text_type(exc))
+            raise PluginError('Invalid project value, numbers only')
```
jamtur01: Is this descriptive enough? The msg confuses me. :)

jamtur01:
Minor comments otherwise LGTM but mostly via reference to other Sentry plugins. I think those folks will need to weigh in.

src/sentry_plugins/clubhouse/plugin.py
miketheman: Probably not.  It was the last thing I did, and I don't have a good test case for it either.
I think a more evolved plugin could perform a `list` of projects, allow the user to type in a name, and then retrieve and store the `id`, preventing the user from seeing numbers at all.
But for now, a better error message might be:

> Invalid Project ID. Project IDs are numbers-only, and can be found on the Project's page

Trying to keep it concise.

jamtur01: +1 to that message. 

Yeah I started to add "this needs a test" but... :(

markkevans:
I would love to see something like this available

We are currently doing this via a google cloud function, some of the things we'e found useful are

- Per sentry project setting which clubhouse project the issue should be assigned to (looks like this is done)
- Per sentry project setting which clubhouse workflow state the issue should be assigned to
- Adding a comment to the existing story in clubhouse when the same event happens multiple times

We do this using a webhook so is automatic, the above looks like you'd need to click the "create clubhouse issue" button for it to appear?

src/sentry_plugins/clubhouse/client.py
```diff
@@ -0,0 +1,24 @@
...
+    #     return self._request(*args, **kwargs)
+
+    def get_story(self, story_id, token):
+        return self.get('/stories/%s' % story_id, params={'token': token})
+
+    def create_story(self, project, data, token):
+        story_data = {
+            'project_id': project,
+            'name': data['title'],
+            'description': data['description'],
```
miketheman: Discuss: Should this automatically add `'story_type': 'bug'` to Stories created by users from Sentry?  Currently defaults to `feature`.

jamtur01: +1

jamtur01:
@markkevans I presume it has a throttle so you don't end up with a million comments?

src/sentry_plugins/clubhouse/plugin.py
dcramer: I never got around to finishing this:

https://github.com/getsentry/sentry/pull/6250

markkevans:
@jamtur01 Good point, we don't get that many alerts so it's not been an issue, we have also configured sentry to send alerts at most every 30 minutes which should reduce the need for the cloud function to need to deal with rate limiting

src/sentry_plugins/clubhouse/plugin.py
miketheman: @eliperkins Thanks - I considered that, and since I _think_ the org name can be altered, that would make this connection more brittle.

@dcramer Do you think you might get back to it soon, or would this be an acceptable first implementation for this integration?

My thoughts are that since the API call will only occur when loading a Sentry issue that is _already linked_ to a Clubhouse Story, the volume will be minimal, and could be shipped sooner.
I could add a comment with a link to that issue, to revisit when the core sentry change is implemented.

src/sentry_plugins/clubhouse/client.py
miketheman: Great! I'll implement the change.

src/sentry_plugins/clubhouse/plugin.py
miketheman: @eliperkins Thanks - I might be able to use that endpoint, I haven't fully explored it yet.

Are you familiar with the "Link to existing issue" behavior? It can be seen with the GitHub/GitLab/Asana integrations, where the user can type in either an issue ID, or some text of the title, and the browser will perform XHR requests to Sentry server, which in turn will run some code against the 3rd party's search mechanism.

I'll try to take another pass at this sometime soon.

Some relevant sections from the GitHub plugin, that will help me find this later:

- https://github.com/getsentry/sentry-plugins/blob/5ea4fbff894bc43d229a3039fb936a2d10eeedcd/src/sentry_plugins/github/plugin.py#L71-L79
- https://github.com/getsentry/sentry-plugins/blob/5ea4fbff894bc43d229a3039fb936a2d10eeedcd/src/sentry_plugins/github/plugin.py#L108-L129
- https://github.com/getsentry/sentry-plugins/blob/5ea4fbff894bc43d229a3039fb936a2d10eeedcd/src/sentry_plugins/github/plugin.py#L197-L220

src/sentry_plugins/clubhouse/client.py
miketheman: Unfortunately, since the above line hasn't changed, GitHub won't collapse this comment stream despite it being implemented.

src/sentry_plugins/clubhouse/plugin.py
eliperkins: Ah, gotcha! Thanks for the link to more info.

One more alternative I could think of would be to use the [`/projects/:id/stories`](https://clubhouse.io/api/rest/v2/#List-Stories) endpoint, since this implementation does link a project ID to the integration currently.

The `/search/stories` endpoint is definitely more flexible and possibly more relevant to the user, as it does ranking based on term frequency, age decay, completion, etc., so it may be a better fit here.

src/sentry_plugins/clubhouse/plugin.py
dcramer: Let me take a look and see. Otherwise you could store the value as something JSON encoded, and then decode it when it comes out (allowing you to store multiple values).

src/sentry_plugins/clubhouse/plugin.py
miketheman: @eliperkins I'll give it a shot, and see what I come up with.

miketheman: @eliperkins The Stories search docs could use some clarity. Is there a good place to discuss that?

miketheman:
# Progress update
I've made some modifications, as can be seen in recent commits. (Happy to squash pre-rebase/merge.)

## Improvements
- Implemented the 'link' concept, using the `search` api, as recommended by @eliperkins . There's a TODO for something to be done about the Clubhouse API, or to make more complex code decisions here.
- Refactored the ClubhouseClient to pass the `token` as part of the params. This seemed like the right way to do it, but there hasn't been too many other plugins with this pattern to compare.


## Remaining
- Storing the return value `app_url` from a create/link operation, [as mentioned by @dcramer](https://github.com/getsentry/sentry-plugins/pull/324#discussion_r171007414). I'm not I fully understand how to store anything about an event in a json encoded format, so any pointers would be helpful.

## Note
Tests have begun failing from a victorops plugin, I don't see how these are related.  @dcramer Did the upstream sentry project change something recently about the backtrace?

## Feedback

Keep it coming! Notes on specific lines in the code is most helpful!

dcramer:
I've landed the change in core that improves the ability to store additional fields. We're going to get this live in prod and make sure its fully stable, but you should be able to build off the API now.

dcramer:
APIs should be stable for the storage changes per the PR. Let me know if there's any questions and I'll try to answer them fairly quickly.

src/sentry_plugins/clubhouse/plugin.py
```diff
@@ -0,0 +1,173 @@
...
+        field = request.GET.get('autocomplete_field')
+        query = request.GET.get('autocomplete_query')
+        if field != 'issue_id' or not query:
+            return Response({'issue_id': []})
+
+        project = self.get_option('project', group.project)
+
+        client = self.get_client(group.project)
+
+        # TODO: Something about the search API won't allow an explicit number search.
```
eliperkins: Do you have more details about this? Story IDs are indexed, and should be boosted given an exact match.

Given some more details, we can tweak this search API to fit the needs or explore a better solution here, perhaps using `GET /stories/:id`.

miketheman: Certainly! It appears that running a call like:

```shell
curl -X GET \
  -H "Content-Type: application/json" \
  -d '{ "query": "project:123 250'" }' \
  -L "https://api.clubhouse.io/api/v2/search/stories?token=$CLUBHOUSE_API_TOKEN"
```
doesn't return Story 250 from project 123 (replace with valid values as per your project/stories).

The [Search Operators](https://help.clubhouse.io/hc/en-us/articles/360000046646-Search-Operators) doc doesn't show any keywords for `story:250` or `story_id:250`

As mentioned in the TODO, we _could_ complicate [this code](https://github.com/getsentry/sentry-plugins/pull/324/files/b826eb76b14a843e2e1b9942e31d0ebc5139e784#diff-53197359e7eaf7e2ac5f7b89978790a1R138) to try and detect if the input is a series of `int` and then run the `GET /stories/:id` call, but it feels like the API should have some mechanism by which the search is done efficiently.

miketheman:
@dcramer Thanks for the heads up! I'll try to figure out the right way to use that API.

src/sentry_plugins/clubhouse/plugin.py
jamtur01: +1

dcramer:
@miketheman take a look at the VSTS plugin with https://github.com/getsentry/sentry-plugins/pull/330 applied -- it should be fairly condensed and obvious from there

src/sentry_plugins/clubhouse/plugin.py
jeremyheiler: @miketheman: Jeremy from Clubhouse here. Story ids are unique per org, so searching by story id and project id doesn't narrow the search any more than searching just by story id. If the query is nothing but the story id, then the first result will be that story. However, if you want only that story, you can query with `id:250`. I hope this helps!

miketheman: @jeremyheiler Thanks! I'll give that a shot, and it works out, great. I still think that I would have to make a distinction in my code to determine if the input is int-only and prepend the search query with `id:`. Is there any possibility of the search API allowing a straight int with no identifier to return the story?

jeremyheiler: @miketheman Curious, what might the value be, other than an int? Generally speaking, I'm not sure we'd want to treat an arbitrary integer in the query as a possible story id. I'll think on it, through.

miketheman: @jeremyheiler As I tried to describe above, and apologies if it wasn't clear, I've implemented a single interface for Sentry to "link to existing stories" - which uses the Sentry built-in autocomplete functionality to pass it to this plugin to handle. 
As the user types, the function fires off a query, in this case to the CH Story Search API, and if the user types in `some project title` then stories matching that will be returned, only if they exist in the configured project (by using the `project:<id>` in the search string behind the scenes.
If the user types in an `int` only, using the **same search function as strings** does not appear to return the story - hence the question of "who should own this logic?"

I can totally implement a regex-based string parser to detect if it's only an ID, and then use the `get_story(id)` method, but that introduces conditional logic to the plugin -- whereas the search API could (and should) be able to return the values for `project:123 250` where 250 is the story id.

miketheman:
@dcramer I used the api as recommended in 204d32b - please take a look and see if I did it right? I was trying to figure out how I could independently determine if the data is stored in the db/redis and didn't find a simple way to decode the `data` column on `sentry_message` table.

dcramer:
@miketheman at a glance looks good. You'd have to shell:

```python
$ sentry shell
from sentry.models import *
queryset = GroupMeta.objects.order_by('-id')[:10]
print '\n'.join('{}={}'.format(meta.key, meta.value) for meta in queryset)
```

miketheman:
@dcramer Thanks - confirmed that what's stored is what I wanted from also looking at the data layer.

```pycon
In [1]: from sentry.models import *
   ...: queryset = GroupMeta.objects.order_by('-id')[:10]
   ...: print '\n'.join('{}={}'.format(meta.key, meta.value) for meta in queryset)
   ...:
clubhouse:issue_title=TypeError: Object [object Object] has no method 'updateFrom'
clubhouse:issue_id=185
clubhouse:issue_url=https://app.clubhouse.io/miketheman/story/185
```

miketheman:
I think this is now ready for final review.

There's still a couple of TODOs - one for testing, that I couldn't see if there's an existing way to validate config other than trying to configure, and the other regarding the search api and using numbers to search.

I'd appreciate another round of review, and then with any directives from @dcramer & friends - should I squash & rebase this now, or wait until review is complete?

src/sentry_plugins/clubhouse/plugin.py
dcramer: For context the goal of this search is that the user can type in the "Issue ID" (story ID it seems in this case?) or a partial match on the "Issue Name"to make it easy to find the referenced item in the other system.

screen shot 2018-03-07 at 10 17 15 am
screen shot 2018-03-07 at 10 17 08 am

tests/clubhouse/test_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+        ) == 'https://app.clubhouse.io/example-org/story/1'
+
+    def test_is_configured(self):
+        assert self.plugin.is_configured(None, self.project) is False
+        self.plugin.set_option('token', '12345678-1234-1234-1234-1234567890AB', self.project)
+        assert self.plugin.is_configured(None, self.project) is False
+        self.plugin.set_option('project', '1234', self.project)
+        assert self.plugin.is_configured(None, self.project) is True
+
+    def test_validate_config(self):
```
dcramer: not sure theres much to help here -- we definitely need to spend more time making these things easier to test

dcramer:
@miketheman you can rebase whenever is convenient (or I can squash+merge when ready)

miketheman:
@dcramer rebased against `master`, squashed down and edited commit message to auto-close original issue.

dcramer:
@miketheman will get this out into prod today (we won't announce it yet, but it also won't be behind a feature switch)

eliperkins:
Awesome job, y'all! :shipit:

jamtur01:
Woot! Thanks @miketheman et al.

dcramer:
This is live and can confirm it appears to be functional. I'll have some followups but can sync on those later once we've gotten our new integrations foundation out.

miketheman:
@dcramer Awesome! Just need some pretty logos from https://github.com/getsentry/sentry/pull/7361

miketheman:
I can confirm this is working on my own account as well. Yay!

PR #330, opened 2018-03-05
ref(vsts): Remove duplicated storage code

dcramer:
This removes code which is now part of Sentry core.


dcramer:
https://github.com/getsentry/sentry/pull/7456 is blocking this

PR #332, opened 2018-03-08
feat(github): Improve webhook capabilities

dcramer:
- Add `update_repository` to provider to re-synchronize webhook
- Add tests for create_repository, update_repository, and delete_repository
- Track webhook_events in repo.config


dcramer:
``update_provider`` would have to be called by hand right now, but we can work on improvements here in the future

PR #334, opened 2018-03-20
ref(github): Improve consistency of Pull Request changes

dcramer:


src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -276,8 +277,7 @@ def __call__(self, event, organization=None):
 class PullRequestEventWebhook(Webhook):
     # https://developer.github.com/v3/activity/events/types/#pullrequestevent
     def __call__(self, event, organization):
-        action = event['action']
-        # TODO(maxbittker) handle is_apps correctly
+        # TODO(maxbittker) handle is_apps correctly (What does this comment mean?)
```
MaxBittker: basically, we have two providers (`github` and `github_apps`) and this code doesn't work with github apps right now (but neither does the rest of it AFAIK)

src/sentry_plugins/github/endpoints/webhook.py
```diff
@@ -295,20 +295,17 @@ def __call__(self, event, organization):
...
         title = pull_request['title']
         body = pull_request['body']
         user = pull_request['user']
-        merge_commit_sha = pull_request['merge_commit_sha']
+        # The value of the merge_commit_sha attribute changes depending on the state of the pull request. Before a pull request is merged, the merge_commit_sha attribute holds the SHA of the test merge commit. After a pull request is merged, the attribute changes depending on how the pull request was merged:
+        # - If the pull request was merged as a merge commit, the attribute represents the SHA of the merge commit.
+        # - If the pull request was merged via a squash, the attribute represents the SHA of the squashed commit on the base branch.
+        # - If the pull request was rebased, the attribute represents the commit that the base branch was updated to.
+        # https://developer.github.com/v3/pulls/#get-a-single-pull-request
+        merge_commit_sha = pull_request['merge_commit_sha'] if pull_request['merged'] else None
```
MaxBittker: could you help me understand why this improves the behavior?

MaxBittker: I think we want to create a resolved_in link when the PR is opened, not when the pr is closed, 
(and update that link if the merged_sha changes)

is my understanding correct that this would be None until the merge happens? Does that stil mark the issue as resolved or create an activity item?

dcramer: @MaxBittker this field i think 1) isnt filled except on close, 2) wont be correct until close

dcramer:
@MaxBittker thats ok when you consider this in isolation, but not if you consider that "this needs to represent the actual SHA of a commit that fixes the behavior". e.g. this needs to work beyond GitHub because GitHub's isnt the only kind of change request model.

PR #338, opened 2018-03-23
ref: Django 1.8 migration compat

dcramer:


dcramer:
https://github.com/getsentry/sentry/pull/7263

PR #341, opened 2018-03-26
fix(bitbucket): Update incoming IP ranges

bmunoz89:
Update IP range according to the link https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html#Managewebhooks-trigger_webhookTriggeringwebhooks

Best


src/sentry_plugins/bitbucket/endpoints/webhook.py
```diff
@@ -22,7 +22,11 @@
 
 # Bitbucket Cloud IP range:
 # https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html#Managewebhooks-trigger_webhookTriggeringwebhooks
-BITBUCKET_IP_RANGE = ipaddress.ip_network(u'104.192.136.0/21')
+BITBUCKET_IP_RANGE = list(ipaddress.ip_network(u'104.192.136.0/21').hosts()) + [
```
mattrobenolt: Just noting, this is 2048 IP addresses. We should avoid doing this.

MaxBittker:
thanks for opening the PR! I went and fixed this in #341 but I appreciate your help initiating this

PR #346, opened 2018-04-03
fix(slack): Fix unicode formating

EvanPurkhiser:
Fixes: [SENTRY-5SW
](https://sentry.io/share/issue/5335afbf14db4d789111abe00f99927b/)

### DIFF
--- a/src/sentry_plugins/slack/plugin.py
+++ b/src/sentry_plugins/slack/plugin.py
@@ -187,7 +187,7 @@ class SlackPlugin(CorePluginMixin, notify.NotificationPlugin):
                 fields.append(
                     {
                         'title': 'Triggered By',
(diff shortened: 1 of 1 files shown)


src/sentry_plugins/slack/plugin.py
```diff
@@ -189,7 +189,7 @@ def notify(self, notification):
                 fields.append(
                     {
                         'title': 'Triggered By',
-                        'value': ', '.join('<%s | %s>' % r for r in rules),
+                        'value': u', '.join(u'<{} | {}>'.format(*r) for r in rules),
```
EvanPurkhiser: `r` is a two item tuple setup here

https://github.com/getsentry/sentry-plugins/blob/7a6824b75860faebc461c20db6c90b40ff0111e0/src/sentry_plugins/slack/plugin.py#L186

PR #351, opened 2018-04-19
fix(bitbucket): commit file changes

MaxBittker:
fixes (APP-70)


src/sentry_plugins/bitbucket/client.py
```diff
@@ -100,13 +100,15 @@ def transform_patchset(self, patch_set):
...
-        diff_file = self.get(
+        resp = self.get(
             '/2.0/repositories/{}/diff/{}'.format(
                 repo,
                 sha,
             ),
             allow_text=True,
         )
+
+        diff_file = resp.text
```
MaxBittker: this was the problem 😬

PR #354, opened 2018-04-27
feat(vsts): commit file changes (wip)

MaxBittker:
Adds commit file change information to vsts commit data by calling the vsts API


src/sentry_plugins/vsts/client.py
```diff
@@ -13,12 +13,12 @@
...
 
 class VstsClient(AuthApiClient):
-    api_version = '3.0'
+    api_version = '4.1'
 
     def request(self, method, path, data=None, params=None):
         headers = {
             'Accept': 'application/json; api-version={}'.format(self.api_version),
-            'Content-Type': 'application/json-patch+json',
+            'Content-Type': 'application/json',
```
MaxBittker: :-|

dcramer: Are you sure this doesn’t break things? It was required before.

MaxBittker: do you remember why? it was causing the commit fetching requests to 500 for me 😬

dcramer: Their API required it for PATCH requests. If you prove it still works with any call that does PATCH within this patch then that'd satisfy my concern. It's possible the new version of the API is fine now without it.

src/sentry_plugins/vsts/repository_provider.py
```diff
@@ -71,26 +71,58 @@ def create_repository(self, organization, data, actor=None):
...
+            'add': 'A',
+            'delete': 'D',
+            'edit': 'M',
+        }
+        file_changes = []
+        # https://docs.microsoft.com/en-us/rest/api/vsts/git/commits/get%20changes#versioncontrolchangetype
+        for change in patch_set:
+            change_type = type_mapping.get(change['changeType'])
+
+            if(change_type and change['item'] and change['item']['gitObjectType'] == 'blob'):
```
macqueen: should the second one be `change.get('item')`?

src/sentry_plugins/vsts/repository_provider.py
```diff
@@ -71,26 +71,58 @@ def create_repository(self, organization, data, actor=None):
...
+        n = 0
+        for commit in commit_list:
+            commit.update(
+                {'patch_set': self.transform_changes(
+                    client.get_commit_filechanges(
+                        repo.config['instance'], repo.external_id, commit['commitId'])
+                )})
+
+            n += 1
+            if n > 90:
```
macqueen: small nit, i think it's nice to put something like this in a constant like `MAX_FILE_CHANGES` or something so that when we inevitably are looking back three months from now to see if there is a limit because a customer has asked, it's easy to find

src/sentry_plugins/vsts/repository_provider.py
MaxBittker: good call

src/sentry_plugins/vsts/repository_provider.py
MaxBittker: true!

src/sentry_plugins/vsts/client.py
```diff
@@ -102,6 +102,7 @@ def update_work_item(self, instance, id, title=UNSET, description=UNSET, link=UN
                 id,
             ),
             data=data,
+            patch=True
```
dcramer: you already know its patch because thats the method name

PR #367, opened 2018-06-26
(fix)Add support for emergency notifications in Pushover integration

ceorourke:


src/sentry_plugins/pushover/plugin.py
```diff
@@ -30,6 +35,7 @@ def get_config(self, **kwargs):
         )
 
         apikey_field.update({'name': 'apikey', 'label': 'API Key'})
+
         return [
             userkey_field, apikey_field, {
                 'name':
```
lauryndbrown: No idea why this was originally formatted this way. Care to correct it? Let's not continue that trend. So, if you have:
```python
{
'name':
'retry',
'label':
'Retry',
'type':
'number',
'required':                
False,
}
```
It should actually look like:
```python
{
'name': 'retry',
'label': 'Retry',
'type': 'number',
'required': False,
}
```

Typically, in larger codebases there's a code formatter. I don't remember if there's in this repository.

src/sentry_plugins/pushover/plugin.py
```diff
@@ -49,9 +55,40 @@ def get_config(self, **kwargs):
...
+                False,
+                'placeholder': 'e.g. 9000',
+                'help': 'How many seconds your notification will continue to be retried for. Maximum of 10800 seconds. Only required for "Emergency" level priority.'
+            },
         ]
 
+    def validate_config(self, project, config, actor):
+        if int(config['priority']) == 2:
+            try:
+                assert config['retry'] >= 30
```
lauryndbrown: I see what you're doing here, and this will work totally fine. But I think a better way would be:
```python
def validate_config(self, project, config, actor):
   if int(config['priority']) == 2 and config['retry'] < 30:
     self.logger.exception(six.text_type('Retry not 30 or higher.'))
     raise PluginError('Retry must be 30 or higher.')
return config
```

Also, does `config['retry']` need to be converted at all?
And I assume you probably don't want just `Assertion error` logged.

src/sentry_plugins/pushover/plugin.py
mattrobenolt: The old code formatter did this.

https://github.com/getsentry/sentry-plugins/commit/b54b6dcfc1300fb8d9087a6f849a15a60e486259

lauryndbrown: Man.... that's really weird. Looks like it made a mistake. I suppose because the `choices` thing.

mattrobenolt: I mean, I'm not going to defend it. :) Just pointing out that a human did not write code that looked like that.

PR #368, opened 2018-07-03
Add custom message option to Slack plugin

Dazix:


EvanPurkhiser:
Hi @Dazix! Thanks for the pull request! I've assigned this to @saragilford our product manager to make the decision if we want to merge this in for our legacy slack plugin :) The code looks great!

The new slack plugin unfortunately doesn't offer any kind of configuration like this, however it is the integration that is supported now, sometime in the future we would like to deprecate this plugin.

@saragilford any thoughts from your end on merging this?

saragilford:
Hey @Dazix ! Appreciate your contribution :) Would you mind posting a screenshot of what this would look like?

Dazix:
Integration details page
Result in slack

src/sentry_plugins/slack/plugin.py
```diff
@@ -69,6 +69,13 @@ def get_config(self, project, **kwargs):
                 'placeholder': 'e.g. #engineering',
                 'required': False,
                 'help': 'Optional #channel name or @user',
+            }, {
+                'name': 'custom_message',
+                'label': 'Custom message',
```
saragilford: Can `Message` be capitalized here?

src/sentry_plugins/slack/plugin.py
```diff
@@ -69,6 +69,13 @@ def get_config(self, project, **kwargs):
...
+                'type': 'string',
+                'placeholder': 'e.g. Hey <!everyone> there is something wrong',
+                'required': False,
+                'help': 'Optional e.g. <!channel> for send notification to everyone in channel',
```
saragilford: Since this legacy Slack integration only sends alerts to one channel, most of our users will have too high of a volume of events to find sending a notification for every exception helpful. While we should definitely keep this functionality in for users like you for whom this is helpful, can we use the following help text instead?

`Optional -  Slack message formatting can be used`

src/sentry_plugins/slack/plugin.py
Dazix: Ok, why not.

src/sentry_plugins/slack/plugin.py
Dazix: Of course.

EvanPurkhiser:
Have to figure out what's wrong with percy builds :/

Dazix:
@EvanPurkhiser Is there anything I can help with?

PR #370, opened 2018-07-10
Update README.md

Smolevich:
In section `Gitalb` I add information about token, right way to use [personal access tokens](https://docs.gitlab.com/ce/api/README.html#personal-access-tokens), not [access token](https://docs.gitlab.com/ce/api/README.html#oauth2-tokens). If in input texrt field `Access Token` of interface user types `access_token`  instead `Personal access token` he catch error `Unauthorized: either your access token was invalid or you do not have access`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,10 +63,9 @@ GitLab
 ------
 
 Go to your project's configuration page (Projects -> [Project] -> Issue Tracking) and select
-GitLab. Enter the required credentials and click save changes.
-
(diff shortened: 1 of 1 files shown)


EvanPurkhiser:
Hi @Smolevich! Thanks for the PR. For some reason our percy is broken right now so I can't merge this, but I went ahead and made a seperate PR (#375) to add his.

Thank you again for the contribution!

PR #377, opened 2018-07-18
lint: bring back sentry.lint.sentry_check:SentryCheck

JoshuaRLi:
ref: https://github.com/getsentry/sentry/pull/9053


JoshuaRLi:
Looks like getsentry/sentry#9053 needs to be merged in before CI can pass here.

PR #382, opened 2018-07-20
feat(trello): Add trello plugin (WIP)

billyvg:
lots of copy and paste...

This is an update of an old plugin, it's currently broken in prod because of a select2 removal.


src/sentry_plugins/trello/plugin.py
```diff
@@ -0,0 +1,260 @@
...
+        token_value = get_from_initial(initial, 'token')
+
+        if token_value:
+            token['has_saved_value'] = True
+            token['prefix'] = token_value[:6]
+            token['required'] = False
+
+        config = [key, token]
+
+        if key_value and token_value:
```
lauryndbrown: @billyvg I have some fears this will not work. The idea way back when was to create a "dynamic" Django form. So, if `[key, token]` are present try to get the organizations. Wrote a lot of this to work around the fact that the move away from Django forms broke Trello. I don't know what is on your plate right now, or if it's even worth making a more permanent fix here, but this really needed a react solution, and I was making due with the skills I had.

lauryndbrown (changes requested):
@billyvg can you ensure by testing (even if just manually) that Trello still works. Trello is pretty easy to set up and shouldn't be difficult to see if it works.

src/sentry_plugins/trello/plugin.py
```diff
@@ -0,0 +1,260 @@
...
+                'name': 'board',
+                'label': 'Board',
+                'type': 'select',
+                'choices': board_choices,
+                'readonly': False,
+                'required': True,
+            },
+            {
+                'name': 'list',
+                'depends': ['board'],
```
billyvg: So this is the solution I came up with @lauryndbrown for dynamic forms ^ + https://github.com/getsentry/sentry/pull/9136

billyvg (commented):
This was my attempt to get Trello working because right now it's broken due to a library we replaced. However, other older plugins are similarly affected, so I did a different workaround: https://github.com/getsentry/sentry/pull/9168/files

I'm not sure if I thoroughly tested the initial configuration but the create issue was tested and worked.

lauryndbrown:
@billyvg I expect it'll probably affect the initial configuration the most. I can actually pull your branch (and the other related one) really quickly and see if it works. Is it in a state that I could do that?

billyvg:
@lauryndbrown Yeah it should be working, but honestly I wouldn't prioritize this if you are working on other things atm. I appreciate it though!

PR #388, opened 2018-08-02
feat(pagerduty): Add tag-based routing

bwester:
Adds a basic UI and mechanism for routing event notifications to
different PagerDuty service keys based on the tags present in the event.
This behavior is useful for handling events from large services where
different teams are responsible for different internal components.

I admit the UI is very basic, but it works. I'm happy to adjust as needed
to make this mesh will with the rest of Sentry, though my understanding
of Sentry/Django UI design is limited.

cc: @mattrobenolt


bwester:
Hi, is there anything I can do move this PR along?

mattrobenolt:
Thanks! 🥇

PR #396, opened 2018-09-11
feat(gitlab): use python-gitlab library for requests 

max-wittig:
[python-gitlab](https://github.com/python-gitlab/python-gitlab) is a very well maintained library.
Advantages of using the library, instead of hand-crafted requests

* Reduce maintenance effort/complexity in sentry for the plugin
* Obey rate-limits given by the GitLab instance (prevent 'too many requests' errors)
* Enable features like getting all members that have access to a project -> b0f4243


mattrobenolt:
It seems we could pass it through the `__init__` of the `GitLab` object: https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/__init__.py#L77

But I'm not sure if this provides much value to us vs just adding support for a new endpoint to support the missing feature vs dealing with potential API compatibilities in the future with a new library.

We greatly favor just making our own API calls when the APIs are simple enough to do so. But I'll leave this decision up to someone else. I just wanted to point out why it's very unsafe for us to do so as-is.

max-wittig:
@mattrobenolt Thanks for the review and sorry about potentially introducing a security hole.
I don't know about that.

python-gitlab does support custom sessions: https://gist.github.com/gpocentek/bd4c3fbf8a6ce226ebddc4aad6b46c0a

I will probably be able to make the required changes tomorrow.

mattrobenolt:
I really don't like this library.

https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/__init__.py#L106-L107

I would highly suggest we don't do this since at a cursorly glance, I'm already concerned.

mattrobenolt:
https://github.com/python-gitlab/python-gitlab/pull/488/files is an even better reason for us to not use this library. This is extremely unsafe code for us and can be a DoS vector.

mattrobenolt:
We absolutely cannot just `time.sleep()` until the rate limit is expired. So we'd have to leave that functionality disabled. Especially since this may be custom servers. What happens if your server sends back `Retry-After: 99999999`?

mattrobenolt:
Oh, or even better, it's also within a `while True`, so a server misbehaving or not, simply needs to send back `Retry-After: 1` forever, and the code will never exit.

We 100% cannot use this library. This is why we write our own light shims over APIs, unfortunately. All code we bring in needs audited, and this one is already pretty glaringly unsafe for us.

Issue #399, opened 2018-09-13
Gitlab Assignee not populating in gitlab issue creation


MattPark:
When your gitlab users are created at the group level, and inherited at the project/repo level, they are not populated in the list of assignees.  Honestly, I suspect that this is how must teams set up gitlab so that you don't have to redo the permissions every time.

![image](https://user-images.githubusercontent.com/1690680/45460315-9a1c6000-b6ca-11e8-8335-a5053bd55724.png)

The issue is confirmed and discussed here: https://gitlab.com/gitlab-org/gitlab-ce/issues/40202#note_100326516

The cause is that sentry is using the incorrect API endpoint to pull the data. 

> We have two APIs for listing project members:
> 
> `GET /projects/:id/members` - This returns only direct members. (We should probably update the docs to make it clear that only direct members are returned.)
> 
> `GET /projects/:id/members/all` - This returns direct as well as inherited members.
>  
> I can confirm that all members including those added from a group are also returned by the `GET /projects/:id/members/all` API. But the docs does not mention this.

I plan to submit a pull request to resolve this issue.

MattPark:
Pull request has been sitting for 3 months, gone stale once already, please merge

PR #400, opened 2018-09-13
Fix assignees to inherit group members

MattPark:
As discussed in #399, this pull request should fix the missing gitlab assignee issue


markstory:
I ended up merging #451 which contains the same changes.

PR #406, opened 2018-10-02
fix(phabricator): Only format unicode strings

mnoble:
These ascii strings weren't able to interpolate unicode bytes causing errors like https://sentry.io/sentry/sentry/issues/627225450/.

Fixes SENTRY-73K


mattrobenolt:
Some day everyone will learn. I should get the linter in though so it actually yells.

I need to fix all of them in getsentry and here first though because I can commit to it.

Issue #317, opened 2018-02-08
Unnecessary info logging message: sentry.plugins.github: apps-not-configured


reedloden:
Noticed that on start-up I see `08:25:18 [INFO] sentry.plugins.github: apps-not-configured` several times (one per worker?). Looks like leftover debug that can be removed.

PR #407, opened 2018-10-03
fix(github): Remove unnecessary log when plugin not configured (#317)

0xdiba:


agis:
Ping!

markstory:
@agis I think this needs to be rebased.

PR #431, opened 2018-12-28
Add more info to victorops integration

mikel-arena:
Along with the other information, also push project id, issue id and details


src/sentry_plugins/victorops/plugin.py
```diff
@@ -82,6 +82,8 @@ def notify_users(self, group, event, fail_silently=False, **kwargs):
             entity_display_name=event.title,
             state_message=self.build_description(event),
             timestamp=int(event.datetime.strftime('%s')),
+            project=event.project.id,
+            issue=event.event_id,
```
markstory: If you're looking for the issue in sentry you would want the `group.id` or `group.qualified_short_id` as those will let you find the issue in sentry more easily.

mikel-arena: In the case of this url: https://sentry.io/arena/production-system/issues/XXXXXXX, what property is the XXXX? Is that `group.id` or `group.qualified_short_id`?

markstory: That would be `group.id`.

mikel-arena: I guess while I'm asking data questions, the "project" in the url, is that `event.project.id`? (Also, even better, are there api docs that I can find what data these objects contain?)

(ill also make that change to `group.id`). Thanks!

mikel-arena: hmm. now that I look at it, we are sending `group.id` as the `entity_id` in victorops (line 81 in plugin.py), but when I look at the victorops alert, that `entity_id` is a GUID and the "issue id" in the url is a 9 digit int.

mikel-arena: so for some back-info, I want to create the link to the sentry issue in victorops in an outgoing webhook, which is why I need project-id, issue-id and so forth, but it looks like the api has a field called `permalink` that IS the url to the issue. Do I have access to this field in `plugin.py`?

ex: `"permalink": "https://sentry.io/the-interstellar-jurisdiction/pump-station/issues/1/", `

markstory: > I guess while I'm asking data questions, the "project" in the url, is that event.project.id

The project in the url is the group's/event's project which can be reached from `event.project` or `group.project`

> Also, even better, are there api docs that I can find what data these objects contain?

I don't think there are any API docs for plugins.

If what you want is to generate a URL back to the sentry issue you can use `group.get_absolute_url()` to get a URL which will point at the issue details page.

mikel-arena: > If what you want is to generate a URL back to the sentry issue you can use group.get_absolute_url() to get a URL which will point at the issue details page.

this is exactly what I want! Updating PR now. Thanks for the quick responses!

mikel-arena: updated

mikel-arena:
@markstory thanks for approving. Do you have any idea the timeline before this gets merged and deployed? We're looking to update our Victorops alarms asap to stop duplicate sentry/victorops alerts.

dcramer:
Thanks!

PR #434, opened 2019-01-04
feat(splunk): Support security header and user interfaces on events

dcramer:


setup.py
```diff
@@ -31,6 +31,7 @@
 ]
 
 install_requires = [
+    'anonymizeip',
```
mattrobenolt: I assume you don't also want this, if you've vendored it.

dcramer:
Also threw in a change in timeout - default was 30 which is too lenient for this kind of process.

PR #445, opened 2019-01-18
(readme) bitbucket: add info about Bitbucket configuration

s-chizhik:
From here:
https://github.com/getsentry/sentry/issues/9205#issuecomment-428165631


s-chizhik:
pew pew?

PR #448, opened 2019-02-28
fix(jira) Stop erroring on expired signatures

markstory:
When user's jira integrations have expired signatures there is little to nothing we can do. Stop tracking these as errors in sentry when they are just background noise.

Fixes SENTRY-1Z0


mattrobenolt:
omg this is the best thing ever!

EvanPurkhiser:
You are a god mark

PR #467, opened 2019-04-23
refs: Remove hipchat plugin

wedamija:
Hipchat cloud has shut down. We have 4 orgs that are using a self hosted Hipchat. Out of these, only
one is paid, on s1. Support is going to contact them just to give them a heads up, but it's not
worth us spending time maintaining this legacy integration, and it also makes a bunch of noise in
our Sentry project.

Fixes SENTRY-9QA
Fixes SENTRY-8G7
Fixes SENTRY-A5T

Will follow up with a separate PR to drop the hipchat tables.


wedamija:
Yeah, I haven't gotten around to it yet, but there are a few things that that I think need to be removed first.

PR #478, opened 2019-05-13
chore: Add precommit hooks - black/isort

EvanPurkhiser:
- Applys black (two commits, one for general formatting, one for quote changes)
- Applys isort fixes
- Adds precommit configuration.


src/sentry_plugins/amazon_sqs/plugin.py
```diff
@@ -4,7 +4,6 @@
 
 from sentry.plugins.bases.data_forwarding import DataForwardingPlugin
 from sentry.utils import json
-
```
EvanPurkhiser: Unfortunately there's no way to specify separate 1st party module groups. I think this is fair to group sentry and sentry_plugins though.

mattrobenolt:
Can we do this also with black at the same time? Otherwise, we're going to have 2 different mass reformats.

EvanPurkhiser:
Sure. I'll apply black. I'll integrate the precommit hook as well in that case.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - repo: https://github.com/ambv/black
+    rev: 19.3b0
+    hooks:
+      - id: black
+        python_version: python3.7
+  - repo: https://github.com/pre-commit/mirrors-isort
+    rev: v4.3.19
+    hooks:
+      - id: isort
+        python_version: python3.7
```
EvanPurkhiser: This tool will run with python2.7, but when the next major release of isort happens it will no longer support 2, so this is somewhat pre-emptive.

EvanPurkhiser:
Did some testing in my locale environment. Everything seems to be working fine as far as I can tell. Think this should be good, wcgw

.pre-commit-config.yaml
joshuarli: Consider using https://github.com/asottile/reorder_python_imports, it's better than isort and is a pc hook.

PR #484, opened 2019-06-14
fix: Don't error in amazon_sqs when an org's token has become invalid

wedamija:
We can't resolve these issues, so no need to error to Sentry. Just log to kibana and datadog and
exit.

Fixes SENTRY-AVT


src/sentry_plugins/amazon_sqs/plugin.py
```diff
@@ -61,16 +68,41 @@ def forward_event(self, event, payload):
...
+                aws_access_key_id=access_key,
+                aws_secret_access_key=secret_key,
+                region_name=region,
+            )
+            client.send_message(
+                QueueUrl=queue_url,
+                MessageBody=message,
+            )
+        except ClientError as e:
+            print e.message
```
mattrobenolt: Prob wanna remove this.

wedamija: Damb

PR #490, opened 2019-06-21
py3: Update boto3 version

mikeroll:
Following the boto3 version bump in getsentry/sentry#13777, also bump it here for consistency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ tests_require = [
 install_requires = [
     'BeautifulSoup>=3.2.1',
     # sentry also requires this, so we're just enforcing that it needs to exist
-    'boto3>=1.4.4,<1.5.0',
+    'boto3>=1.4.4,<1.10.0',
     'cached-property',
     'mistune',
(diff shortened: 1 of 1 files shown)


mikeroll:
@mattrobenolt re your concern about the https://github.com/getsentry/sentry-plugins/blob/3dbfd39c8931a39563dfe2a964c3aeda2573bd99/src/sentry_plugins/amazon_sqs/plugin.py#L83

The exception msg did not change:
https://github.com/boto/botocore/blob/1.12.174/botocore/exceptions.py#L391

A manual check with `boto3==1.9.174, botocore==1.12.174` results in a `botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling ...`.

tlebrize:
Do we have any ETA on when this will be merged ? Can i help ?
I need this patch enables amazon SQS on eu-west-3

markstory:
@tlebrize I don't right now. I've been swamped lately and this is a hard to verify change as I'm not actively familiar with the nuances around SQS and S3.

PR #496, opened 2019-07-26
chore: blacken

joshuarli:
Supersedes #495. Main difference is that it's pure black via precommit, no flake8/isort; flake8 and precommit-native hooks should be standardized across all python repos, and there hasn't been any decision regarding that yet.

See also https://github.com/getsentry/getsentry/pull/3074.


joshuarli:
Selenium's timing out in acceptance tests, unsure why or how to fix.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
EvanPurkhiser: Will we add the rest of our pre-commit later? Also is there a way to share pre-commit files?

Makefile
```diff
@@ -36,6 +36,7 @@ clean:
 
 lint: lint-js lint-python
 
+# TODO (joshuarli): move flake8 to pre-commit, keep backwards compat for lint-python
 lint-python:
```
EvanPurkhiser: Ideally we should change all our lint-python stuff to run pre-commit.

joshuarli: That's not semantically correct though. lint-python should be kept, but it should only run flake8 via pre-commit.

EvanPurkhiser: actually. We need this plugin

https://pypi.org/project/flake8-black/

We do something similar for our eslint in javascript.

.pre-commit-config.yaml
joshuarli: Yes.

> no flake8/isort; flake8 and precommit-native hooks should be standardized across all python repos, and there hasn't been any decision regarding that yet.

And if by sharing you mean automagically cross-repo across the network, not that I know of. The config file has to be manually updated. This goes back to my quote above (why I'm hesitant about defining more pre-commit stuff)

joshuarli:
re: acceptance test failures, it happens when webpack doesn't run and fails to templatize a `{version}` in an asset uri (404s during `pytest --no-headless=1`). Fixed in https://github.com/getsentry/sentry/pull/14178.

EvanPurkhiser:
no idea why it fails to templatize if the assets are missing

Makefile
joshuarli: see discussion here: https://github.com/getsentry/getsentry/pull/3087#discussion_r308489363

joshuarli:
Decoupling the pre-commits hooks in ci commits from this PR into #498.

joshuarli:
Added `--install-hooks` to `pre-commit install` so any surprises occur sooner.

Looks like `make install-tests` eventually `make setup-git`'s. So I have to do the pyenv stuff to get `python3` on path for pre-commit to install black. CI isn't run though, that part is still in #498. This isn't the case with https://github.com/getsentry/getsentry/pull/3087 since the make recipes for CI are a bit different (and rightly decoupled from development stuff).

.travis.yml
```diff
@@ -35,7 +35,11 @@ matrix:
         - 'export PATH=$PATH:~/.bin'
         - nvm install $TRAVIS_NODE_VERSION
         - npm install -g "yarn@${YARN_VERSION}"
+        # trusty comes with pyenv-managed 2.7 and 3.6 - we need python3 -> 3.6 on the PATH for pre-commit to install black during 2.7 tests.
+        - pyenv global 3.6.3 2.7.14
```
EvanPurkhiser: Why not 3.7?

joshuarli: IIRC I shelled into travis and these were the highest versions available. Can recheck when I bump the ubuntu version in a future pr, 3.7 is definitely preferable.

PR #498, opened 2019-07-30
ci(travis): add pre-commit hooks as a ci step

joshuarli:
Blocked by merge of #496, which this is branched off of. Also wait until people have pre-commit in their workflow.


joshuarli:
Pre-emptively closing this since sentry-plugins will be moved into sentry soon.

PR #504, opened 2019-08-21
make: disable pep517 and pin pip to 19.1.1

joshuarli:
Forgot to do this. Also semi-related to the pinning of pip: https://github.com/getsentry/sentry/pull/14463


Makefile
```diff
@@ -1,27 +1,32 @@
...
+
 install-yarn:
 	@echo "--> Installing Node dependencies"
 	@hash yarn 2> /dev/null || npm install -g yarn
 	# Use NODE_ENV=development so that yarn installs both dependencies + devDependencies
 	NODE_ENV=development yarn install --ignore-optional
 
 install-tests: develop
-	pip install .[tests]
+	pip install ".[tests]" $(PIP_OPTS)
```
BYK:
```suggestion
	pip install $(PIP_OPTS) ".[tests]" 
```
Putting options as early as possible is better to resolve the ambiguity. We may even pull them before the `install` command if supported to prevent any "accidental" package names sneaking into these options.

Makefile
```diff
@@ -1,27 +1,32 @@
...
 
 setup-git:
 	@echo "--> Installing git hooks"
 	git config branch.autosetuprebase always
 	git config core.ignorecase false
 	cd .git/hooks && ln -sf ../../config/hooks/* ./
-	pip install "pre-commit>=1.10.1,<1.11.0"
+	pip install "pre-commit>=1.10.1,<1.11.0" $(PIP_OPTS)
```
BYK:
```suggestion
	pip install $(PIP_OPTS) "pre-commit>=1.10.1,<1.11.0"
```

See above.

Makefile
```diff
@@ -1,27 +1,32 @@
 SENTRY_PATH := `python -c 'import sentry; print sentry.__file__.rsplit("/", 3)[0]'`
 UNAME_S := $(shell uname -s)
 
-develop: setup-git install-yarn
-	pip install "pip>=9,<10"
-	pip install -e git+https://github.com/getsentry/sentry.git#egg=sentry[dev,optional]
-	pip install -e ".[tests]"
+PIP_VERSION := 19.1.1
```
BYK: Latest is `19.2.1` so why not that one?

joshuarli: It's simply because we have 19.1.1 in a lot of other places.

I don't know if it's common to do this, but I think we should serve some highly available strings on the web like `http://chromedriver.storage.googleapis.com/LATEST_RELEASE_77`. Something like `https://cfg.sentry.io/PYTHON_PIP_VERSION`. That way, it's not nearly as painful to bump/change stuff like this across all our repositories, and we have an easy-to-find single source of truth. cc @mattrobenolt

Makefile
joshuarli: I agree. The only reason this is here is because this is just how it is currently done in getsentry/sentry.

Makefile
```diff
@@ -1,27 +1,32 @@
...
+PIP_OPTS := --no-use-pep517 --disable-pip-version-check
+
+develop: ensure-pip setup-git install-yarn
+	pip install -e git+https://github.com/getsentry/sentry.git#egg=sentry[dev,optional] $(PIP_OPTS)
+	pip install -e ".[tests]" $(PIP_OPTS)
 	yarn install
 
+ensure-pip:
+	pip install "pip==$(PIP_VERSION)"
```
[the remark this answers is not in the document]
joshuarli: Ok, I read your suggested changes bottom-to-top. We just have a mixture of `python -m pip` and `pip`, and I agree that the former is better. Not so much for the windows reason, but because it ties the pip invocation to the correctness of what python is resolved to (pip could be anything.)

Going forward, since we both seem to have strong opinions, we (at least us) should collaborate on an initial internal draft of best practices like this, then gradually refactor everything over. Not saying we should go to the level of linting for this kind of stuff, but at least get all these opinions in 1 place.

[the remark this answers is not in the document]
BYK: >Going forward, since we both seem to have strong opinions, we (at least us) should collaborate on an initial internal draft of best practices like this, then gradually refactor everything over. Not saying we should go to the level of linting for this kind of stuff, but at least get all these opinions in 1 place.

Agree with the second part but I don't have a strong opinion on any of these (would have requested changes if I did) :D

These are mere highlights or recommendations but as long as your version works and it is not "terrible", go ahead and merge :)

Makefile
BYK: Consistency for consistency's sake, especially for poor practices is not a good thing. Since this is not very important and dangerous, I think it can stay that way but I'd just beware of this consistency trap.

Makefile
BYK: That's an interesting idea. We may look into this. A better thing would be to consolidate all these separate builds and make files under one roof tho as there are more things to diverge.

Makefile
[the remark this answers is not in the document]
joshuarli: Well, the opportunity to escape the consistency trap you mentioned elsewhere is already here, so I'll just apply your suggestions to this pr.

joshuarli:
Superseded by https://github.com/getsentry/sentry-plugins/pull/504.

BYK:
>Superseded by #504.

Did you mean __recursion__?

joshuarli:
> > Superseded by #504.
> 
> Did you mean _recursion_?

Haha oops, totally a copy paste fail. Dunno what it was superseded by but should be easy to find out if someone wants to.

PR #506, opened 2019-08-29
Don't cast Asana workspace ID to int

sgrimm-sg:
Asana's API expects workspace IDs to be passed as strings, not numbers, and it has recently started enforcing that expectation. This causes task creation from the Sentry UI to fail with an error, "Error Communicating with Asana (HTTP 400): workspace: Not a valid GID type: number".


sgrimm-sg:
What does the schedule look like for rolling this kind of change out to sentry.io?

markstory:
@sgrimm-sg It should have gone out 5 days ago.

okdewit:
@markstory I can confirm it's not online on the paid hosted service yet:

image

okdewit:
@markstory I'm not very familiar with Python 2.x development, but I made a PR which might resolve the remaining issue:
https://github.com/getsentry/sentry-plugins/pull/508
The build is currently failing, although it doesn't seem to fail on the changes in the PR.

PR #508, opened 2019-09-06
Convert Asana GIDs to strings

okdewit:
Convert GIDs to strings for Asana API compatibility.

See: https://forum.asana.com/t/asana-is-moving-to-string-ids-updated-with-revised-timeline/29340

### DIFF
--- a/src/sentry_plugins/asana/client.py
+++ b/src/sentry_plugins/asana/client.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 from sentry_plugins.client import AuthApiClient
+from six import text_type
 
 
 class AsanaClient(AuthApiClient):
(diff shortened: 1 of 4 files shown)


okdewit:
The build is currently failing, although it doesn't seem to fail on the changes in the PR.

okdewit:
@markstory I'm attempting to get this through CI, but I'm a bit puzzled that it keeps failing on Django framework deprecation warnings.

Could you advise me how I get this into mergable state? :)

src/sentry_plugins/asana/client.py
```diff
@@ -14,12 +14,12 @@ def get_issue(self, issue_id):
         return self.get("/tasks/%s" % issue_id)
 
     def create_issue(self, workspace, data):
-        asana_data = {"name": data["title"], "notes": data["description"], "workspace": workspace}
+        asana_data = {"name": data["title"], "notes": data["description"], "workspace": str(workspace)}
         if data.get("project"):
-            asana_data["projects"] = [data["project"]]
+            asana_data["projects"] = [str(data["project"])]
```
markstory: You should use `six.text_type()` instead of `str()`. `str()` will throw an exception in python2 when non-ascii bytes are encountered.

markstory:
@sgrimm-sg The master build passed on its last run.

okdewit:
@markstory @sggrimm-sg I did a rebase, which adds the changes from lynnagara to increase CI memory. Now Travis passes.

PR #518, opened 2019-10-02
ref: general spruce-up (make, ci, dev environment)

joshuarli:
Upgrading and fixing dev and ci environment to prepare for djangorestframework upgrades.

- sentry-plugins doesn't explicitly pin drf, it just assumes the subdependency from sentry.
- `make develop` doesn't pull in local sentry as editable, it just pulls upstream master. drf needs to be updated in lockstep with sentry, getsentry and sentry-plugins.
- generally outdated stuff (ci platform, linters, pip version)


Makefile
```diff
@@ -1,30 +1,45 @@
...
 	@echo ""
 
+install-yarn:
+	@echo "--> Installing Node dependencies"
+	@hash yarn 2> /dev/null || npm install -g yarn
+
+install-requirements: install-pip install-yarn
+	yarn install
+	# sentry optional might be unnecessary?
+	$(PIP) install $(PIP_OPTS) 'git+https://github.com/getsentry/sentry.git#egg=sentry[optional]'
```
markstory: Will this also compile with webpack? Do you need that?

joshuarli: I'm not sure, don't really know anything about how js deps with plugins work. Will circle back to this though.

joshuarli: dang mark, you predicted the future https://github.com/getsentry/sentry-plugins/pull/518/commits/c36c198feab2876d22113277c23d31b0b7e55db5

Makefile
```diff
@@ -1,30 +1,43 @@
...
+	$(PIP) install $(PIP_OPTS) 'git+https://github.com/getsentry/sentry.git#egg=sentry'
+	$(PIP) install $(PIP_OPTS) .
+
+install-develop-requirements: setup-git install-pip install-yarn
+	NODE_ENV=development yarn install --ignore-optional
+	$(PIP) install $(PIP_OPTS) -e "../sentry[dev]"
+	$(PIP) install $(PIP_OPTS) -e ".[tests]"
+
+# this is for ci, so no relative sentry
+# XXX: sentry and itself is installed editable to prevent wheel building for now until i figure out why it fails
```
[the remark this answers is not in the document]
joshuarli: Yeah, but it should really be non-editable. We do editable installs in a lot of places when the more correct thing to be doing is not doing that.

joshuarli:
Testing on https://github.com/getsentry/getsentry/pull/3182.

PR #522, opened 2019-10-07
py3(drf): bump djangorestframework to 3.4.x

joshuarli:
Syncing with https://github.com/getsentry/sentry/pull/14974.


joshuarli:
The failing heroku plugin test doesn't seem to be related to this PR, but I'm not sure what the cause is yet.

PR #523, opened 2019-10-08
py3(pytest): bump pytest and friends to last 2.7 supported versions

joshuarli:
Syncing with https://github.com/getsentry/sentry/pull/14997.


joshuarli:
Tests are failing because sentry-plugins pulls sentry master. There currently isn't any way to point it at a specific sentry sha. So either make that happen e.g. a bump-plugins similar to what we have in getsentry, or this is blocked until https://github.com/getsentry/sentry/pull/14997 gets merged.

PR #524, opened 2019-10-10
py3(django): fix Django 1.9 deprecation warnings

joshuarli:
Syncing with https://github.com/getsentry/sentry/pull/15025.


joshuarli:
As far as I've seen, these are the only Django 1.9 changes needed in sentry-plugins. (There's a lot of 1.10, but that's not in scope.)

Most of the Django 1.9 changes are in https://github.com/getsentry/sentry/pull/15025. Interestingly, they were only discovered by running sentry-plugins tests which hit those code paths; sentry-plugins should really be moved into sentry soon.

PR #526, opened 2019-10-17
Add autocreate fatals flag in the sentry-jira plugin

alexef:
The sentry JIRA plugin is useful for us, but this additional flag makes it even more so. We're interested in a JIRA ticket being opened only when the level of the event is "fatal".

Thank you for considering this.


setup.py
```diff
@@ -17,7 +17,7 @@
 
 from sentry.utils.distutils import (BuildAssetsCommand)
 
-VERSION = '9.1.2'
+VERSION = '9.1.3'
```
alexef:
```suggestion
VERSION = '9.1.2'
```

PR #529, opened 2019-10-22
feat: Switch jira_ac to use django migrations

wedamija:
Needed as part of the work to switch sentry to django migrations


markstory (approved):
Looks good. The schema matches what is in prod for this table.

PR #531, opened 2019-10-30
feat: gtfo south

wedamija:
I ran out of gifs


joshuarli:
I approve even though it's already been merged.

PR #538, opened 2019-11-06
migration(plugins): removes all plugins except slack and PD

scefali:
Paired with https://github.com/getsentry/sentry/pull/15468

Must be merged after: https://github.com/getsentry/getsentry/pull/3309


conftest.py
```diff
@@ -11,10 +11,6 @@ def pytest_configure(config):
     # Note: We could manually register/configure INSTALLED_APPS by traversing our entry points
     # or package directories, but this is easier assuming Sentry doesn't change APIs.
     # Note: Order of operations matters here.
-    from sentry.runner.importer import install_plugin_apps
-
-    install_plugin_apps("sentry.apps", settings)
```
joshuarli: Why did this get removed? Do slack/pd not need to be added to INSTALLED_APPS by sentry's importer?

wedamija (approved):
Can you post in #engineering and let people know they'll need to update their sentry-plugins if they pull master and are running getsentry? Just want to make sure we don't break people's devs again

conftest.py
wedamija: Actually yeah, seems like we should keep this since there are still a couple plugins in `sentry.apps`

joshuarli (changes requested):
https://github.com/getsentry/sentry-plugins/pull/538#discussion_r343282061

conftest.py
scefali: @wedamija @joshuarli Hmmm? The only ones in that list were Jira, Jira AC, and SessionStack. All of those are now in sentry-core which has the entrypoint of `sentry.new_apps`.

joshuarli: @scefali Ah you're right, I understand now.

joshuarli (approved):

